PR #22, opened 2017-06-19
fix(userpass): send correct params to passwordReset, update docs

mbroadst:


src/auth/providers.js
```diff
@@ -41,35 +55,66 @@ function userPassProvider(auth) {
...
       fetchArgs.cors = true;
 
       return fetch(`${auth.rootUrl}/local/userpass/reset`, fetchArgs)
         .then(common.checkStatus)
         .then(response => response.json());
     },
 
+
+    /**
+     * Initiate an email signup flow
```
mpobrien: Add: "Will trigger an email to the requested account containing a link with the token and tokenId that must be returned to the server using `emailConfirm()` to activate the account."

PR #27, opened 2017-07-11
Stitch-468 - fix(checkStatus): parse response json on error

mattChiaravalloti:
lmk if you think this is the right direction for adding tests (and if any more tests are actually necessary)

### DIFF
--- a/src/common.js
+++ b/src/common.js
@@ -18,9 +18,13 @@ export const checkStatus = (response) => {
   let error = new Error(response.statusText);
   error.response = response;
 
+  // set error to statusText by default; this will be overwritten when (and if)
+  // the response is successfully parsed into json below
+  error.error = response.statusText;
+
   return response.json()
-    .then(json => Promise.reject(Object.assign(error, json)))
-    .catch(e => Promise.reject(error));
+    .catch(() => Promise.reject(error))
+    .then(json => Promise.reject(Object.assign(error, json)));
 };
 
 export const makeFetchArgs = (method, body) => {
--- a/src/common.js
+++ b/src/common.js
@@ -17,7 +17,10 @@ export const checkStatus = (response) => {
 
   let error = new Error(response.statusText);
   error.response = response;
-  throw error;
+
+  return response.json()
+    .then(json => Promise.reject(Object.assign(error, json)))
+    .catch(e => Promise.reject(error));
 };
 
 export const makeFetchArgs = (method, body) => {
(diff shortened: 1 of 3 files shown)


mattChiaravalloti:
also, that was my bad, including the test-fix branch commits in this branch!

test/test.js
```diff
@@ -130,6 +144,19 @@ describe('Auth', () => {
...
 
+      it('should have an error if local auth is unsuccessful', (done) => {
+        expect.assertions(3);
+        const a = new Auth(null, '/auth');
+        return a.provider('userpass').login('fake-user', 'password')
+          .then(() => done('expected an error on unsuccessful login'))
+          .catch(e => {
+            expect(e).toBeInstanceOf(Error);
+            expect(e.response.status).toBe(401);
+            expect(e.error).toBe('unauthorized');
```
mpobrien: might as well assert something for `errorCode` here too, since it's being set in the response.

src/common.js
```diff
@@ -17,7 +17,10 @@ export const checkStatus = (response) => {
 
   let error = new Error(response.statusText);
   error.response = response;
-  throw error;
+
+  return response.json()
+    .then(json => Promise.reject(Object.assign(error, json)))
```
mpobrien: unless i'm misreading this: isn't this `Promise.reject` just going to immediately trigger the `.catch` on the line after it? Then in that `.catch`, since the promise is rejected with `error` it wouldn't include the json fields that were parsed out in this line.

mpobrien: Actually, now I see what's happening: it *does* trigger the `.catch` but it still works because `Object.assign` modifies its first argument (`error`) directly, so `Promise.reject(error)` will still contain the correct data when it gets there. 
Seems fine, except since we don't use the `e => ...` argument in the subsequent catch we might as well just replace that with `() => ...`.

mpobrien:
I think it would be useful to add an additional test: for cases where the SDK has sent a request that receives an error response that does *not* contain valid JSON (maybe the load balancer has hit an error and we get back an HTML error page, for example). We just need to make sure that some error object is returned without the SDK crashing - it should contain the status code and possibly the statusText, at least.

src/common.js
mattChiaravalloti: good idea. I'm thinking of changing the order:
`response.json().catch(() => Promise.reject(error)).then(json => Promise.reject(Object.assign(error, json)));` which will still work but is a little confusing.

mpobrien: Ah, that's not bad either. I'm fine with either one, as long as the tests agree with it :)

PR #28, opened 2017-07-12
chore(build) remove tarball upload from dist

edaniels:
I don't think we need to do this anymore since we publish to NPM

### DIFF
--- a/.evg.yml
+++ b/.evg.yml
@@ -83,6 +83,7 @@ tasks:
             mkdir tmp
             cd tmp
             npm pack ../
+            mv mongodb-stitch-*.tgz mongodb-stitch.tgz
 
       - command: s3.put
         params:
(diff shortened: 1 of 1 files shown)


mpobrien:
I think there may still be some value in uploading the built tarball somewhere, though. For example, it would give us the ability to `npm install` a tarball from a patch build, or some commit in between the npm releases, and try it out within the admin UI, examples code, etc. I think it will likely come up that we want to test out some version of the SDK in downstream codebases without needing to publish another npm package first since that will put it out into the wild.
Maybe we could just keep this `s3.put` there, but change the remote path to just use `task_id` instead? Though we might also need to add a `mv ./stitch-*.tgz ./stitch.tgz` in the build step so that the changing version number doesn't result in a different file name every time, and the `local_file` argument can refer to a static value.

edaniels:
sgtm

edaniels:
@mpobrien updated

PR #29, opened 2017-07-18
feat(deviceInfo): stitch-118 javascript sdk submits device info upon …

mattChiaravalloti:
…auth

WIP - this is just an idea so far still. No new tests have been written, and the server code has not been updated to look in URL params for device info if it is not in the body.


mattChiaravalloti:
@mpobrien 

files to actually review:
src/auth/index.js,
src/auth/providers.js,
src/client.js,
src/common.js,
src/utils.js,
test/test.js

Again, this might not be "done" yet and certainly isn't tested.

What do you think of the way `getDeviceInfo` handles everything in `src/auth/providers.js`? Do the helper functions I wrote in `src/utils.js` belong there? Nothing special actually needs to be done for storage (or for the `clear` method in the `Auth` class in `src/auth/index.js`) to keep deviceId around--lmk if you think that is not the case.

src/util.js
```diff
@@ -118,4 +120,47 @@ function letMixin(Type) {
...
+function canDeterminePlatform() {
+  return !!platform;
+}
+
+/**
+ * Utility function to get the name of the platform running the code. For a
+ * browser, it will be the browser's name; otherwise, it will be 'node'.
+ */
+function getPlatformName() {
+  return !!platform && platform.name;
```
mpobrien: i don't understand how this line is meant to work - if `platform` is empty, won't this function just return `false` ? since `!!platform` will evaluate to false, and short-circuit the `&&`.

src/util.js
```diff
@@ -118,4 +120,47 @@ function letMixin(Type) {
...
+}
+
+/**
+ * Utility function to get the version of the platform running the code.
+ */
+function getPlatformVersion() {
```
mpobrien: instead of having separate functions for `canDeterminePlatform`, `getPlatformName`, and `getPlatformVersion` maybe we'd be better with just having one function `getPlatform` that returns an object `{name: "...", version: "...."}` or `null` if theres no data available? I may be misunderstanding the intention of `canDeterminePlatform()`...

src/util.js
```diff
@@ -118,4 +120,47 @@ function letMixin(Type) {
...
+/**
+ * Utility function to encode a JSON object into a valid string that can be
+ * inserted in a URI. The object is first stringified, then encoded in base64,
+ * and finally encoded via the builtin encodeURIComponent function.
+ *
+ * @param {Object} obj The object to encode
+ * @returns {String} The encoded object
+ */
+function uriEncodeObject(obj) {
+  return encodeURIComponent(new Buffer(JSON.stringify(obj)).toString('base64'));
```
mpobrien: Sadly `Buffer` won't work in the browser. We may need a base64 encode/decode polyfill to do this in a portable way, such as https://github.com/davidchambers/Base64.js/.

test/test.js
```diff
@@ -249,7 +249,7 @@ describe('anonymous auth', () => {
   beforeEach(() => {
     let count = 0;
     fetchMock.restore();
-    fetchMock.mock(ANON_AUTH_URL, {
+    fetchMock.mock(`begin:${ANON_AUTH_URL}`, {
```
mpobrien: just to confirm the reason for this change: does `ANON_AUTH_URL` by itself just not match if the incoming request url has query params attached to it?

src/common.js
```diff
@@ -10,6 +10,9 @@ export const IMPERSONATION_REAL_USER_AUTH_KEY = '_stitch_impers_real_ua';
 export const USER_AUTH_COOKIE_NAME = 'stitch_ua';
 export const DEFAULT_STITCH_SERVER_URL = 'https://stitch.mongodb.com';
 
+// VERSION is substituted with the package.json version number at build time
+export const SDK_VERSION = VERSION;
```
mpobrien: does `VERSION` get a default value from somewhere, or can we give it one (like "unknown" or something)? as-is it looks like if you were to try to import the module directly from `src`, JS might complain that `VERSION` was never declared, i think.

src/util.js
mattChiaravalloti: you're right; I don't know what I was thinking.

test/test.js
mattChiaravalloti: yes

src/common.js
mattChiaravalloti: This is a good point. If we declare something like `let VERSION = 'unknown'`, I'm pretty sure the plugin will break that by replacing `VERSION` with the version number, making JS complain. If we don't declare it, you're right, then VERSION will be undeclared, making JS complain. I'll think about this more and see what I can do.

mattChiaravalloti:
Ok, I addressed your other comments. I'm going to get the server to start accepting SDK version before adding that line to this code.

src/common.js
```diff
@@ -11,7 +11,11 @@ export const USER_AUTH_COOKIE_NAME = 'stitch_ua';
 export const DEFAULT_STITCH_SERVER_URL = 'https://stitch.mongodb.com';
 
 // VERSION is substituted with the package.json version number at build time
-export const SDK_VERSION = VERSION;
+let version = 'unknown';
+if (typeof VERSION !== 'undefined') {
+  version = VERSION;
+}
+export const SDK_VERSION = version;
```
mpobrien: very clever solution here
😎

mattChiaravalloti:
Added a bunch of tests to `test/test.js` where other auth tests lived. Should I do a refactor to put them in `test/auth.test.js`? Also, should I replicate these tests for each of the other auth forms (beyond just username/password)?

mpobrien:
I wouldn't worry too much about testing this with API key or anonymous logins since they're pretty much the exact same code path as username + password. But, I think a test that checks that device ID in url params is being captured would be valuable, so that we have coverage on `uriEncodeObject`.

package.json
```diff
@@ -27,7 +27,8 @@
     "detect-browser": "^1.7.0",
     "fetch-everywhere": "^1.0.5",
     "mongodb-extjson": "^1.0.5",
-    "query-string": "^4.3.4"
+    "query-string": "^4.3.4",
```
mpobrien: i think you can put these into `devDependencies` since they're only necessary for the tests (for now)

PR #30, opened 2017-07-24
fix(dist): use updated go run command and don't clobber s3 keys

mpobrien:
@edaniels this fixes the tests (was broken by the update to package structure).
This also addresses a similar problem as the one in your other PR for the server repo about the S3 keys. I deleted the branch-based bundle upload since I don't think anyone's relying on it, and they should be using `stable` anyway, so it seems safe to kill off.

### DIFF
--- a/.evg.yml
+++ b/.evg.yml
@@ -49,7 +49,7 @@ functions:
           export STITCH_PATH=$GOPATH/src/github.com/10gen/stitch
           cd stitch-js-sdk/test/bin
           echo "running stitch"
-          go run $STITCH_PATH/main/main.go --configFile ../conf/test_config.json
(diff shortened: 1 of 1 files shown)


.evg.yml
```diff
@@ -90,7 +90,7 @@ tasks:
           aws_key: ${sdks_aws_key}
           aws_secret: ${sdks_aws_secret}
           local_file: stitch-js-sdk/dist/web/stitch.js
-          remote_file: js/library/${revision}/stitch.js
+          remote_file: js/library/${version_id}/stitch.js
```
edaniels: Where are we using these files? Keeping them around for build artifacts?

mpobrien: Yes - kept as build artifacts, and then also copied into `stable` when we want to publish a new release.

PR #31, opened 2017-07-25
Adding simple aggregate to collection.

durran:
Given I'm just getting into Stitch, I wasn't sure of what options would be exposed via the server, such as collations, so I did not include options for the first pass. As for the code, I just followed the patterns that were already in place in `Collection` and in the tests.

### DIFF
--- a/src/services/mongodb/collection.js
+++ b/src/services/mongodb/collection.js
@@ -105,6 +105,15 @@ class Collection {
   }
 
   /**
+   * Executes an aggregation pipeline.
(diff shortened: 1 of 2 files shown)


mpobrien:
👍  Thank you this looks great!
Can you amend the commit message to something like this format?
https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#-git-commit-guidelines
For example:
`feat(mongodb): add simple aggregation to collection`

durran:
@mpobrien Rebased with a squash and updated the commit message.

PR #32, opened 2017-07-25
feat(util): STITCH-482 attach warnings to final result of pipeline requests

mpobrien:
LMK what you all think of this approach - there are some drawbacks to it but it seems to be the simplest thing that doesn't uproot lots of existing code or create backwards API incompatibilities.

(One of the new tests will unfortunately fail since it relies on the corresponding PR in the server repo being merged into master. Can re-run once that's in there.)


mpobrien:
cc @jsflax Would appreciate you weighing on this too, since 482 has implications for the other SDKs as well.

src/util.js
```diff
@@ -1,6 +1,43 @@
...
+        RESULT_METADATA_KEY,
+        {enumerable: false, configurable: false, writable: false, value: metadata}
+      );
+    }
+    return Promise.resolve(res);
+  };
+  const captureMetadata = (data) => {
+    if (typeof data === 'object' && Object.prototype.hasOwnProperty.call(data, RESULT_METADATA_KEY)) {
+      // metadata already attached to this result, pass along
+      metadata = data[RESULT_METADATA_KEY];
```
edaniels: [q] might it be better to not rely on the closure of `metadata` and instead have attachMetadata take the metadata as an argument and produce a `attachMetadata` function?

src/util.js
```diff
@@ -60,16 +97,15 @@ function serviceResponse(service, stages, finalizer) {
       enumerable: false, writable: false, configurable: false,
       value: (resolve, reject) => {
         let result = client.executePipeline(Array.isArray(stages) ? stages : [ stages ]);
-        return (!!finalizer) ?
-          result.then(finalizer).then(resolve, reject) : result.then(resolve, reject);
+        return (!!finalizer) ? result.then(collectMetadata(finalizer)).then(resolve, reject) : result.then(resolve, reject);
```
edaniels: The use of `finalizer`s seems fishy and makes the core logic of `collectMetadata` harder to reason about (checking if we've already attached once). It seems like these are necessary due to the syntactic sugar we have. Might it make sense to have execute pipeline have the `finalizer` logic instead to avoid having to use `collectMetadata` here? Either that or maybe `executePipeline` is really `serviceResponse`.

dist/node/client.js
```diff
@@ -65,13 +65,13 @@ var StitchClient = function () {
       baseUrl = options.baseUrl;
     }
 
-    this.clientAppID = clientAppID;
```
mpobrien: i think we should just ignore it in PRs since it always needs to be freshly built for releases anyway.

edaniels: sgtm. Would cut down on noise too in a PR

mpobrien:
@edaniels good suggestions, thanks - updated. Moved the `finalizer` to be managed entirely within `executePipeline`.

PR #35, opened 2017-08-23
STITCH-676 return value shape of executeNamedPipeline to return just …

mpobrien:
…the result contents


mpobrien:
Dropping this as we've decided to rework this in pipelines v2.

PR #38, opened 2017-09-08
STITCH-600 enforce `npm run build` on publish, remove `dist` from source control

mpobrien:


package.json
```diff
@@ -16,7 +16,8 @@
     "test": "jest",
     "lint": "eslint src/ test/",
     "lintfix": "eslint --fix src/ test/",
-    "publish-docs": "npm run docs && gh-pages -d docs"
+    "publish-docs": "npm run docs && gh-pages -d docs",
+	"prepare": "npm run build"
```
UnicodeSnowman: formatting seems a little funky here

mpobrien: ah, my edit made this a tab instead of spaces - fixed.

PR #39, opened 2017-09-15
STITCH-734 - Adding ability to delete users

haleyowen:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "mongodb-stitch",
-  "version": "0.0.26",
+  "version": "0.0.27",
```
mburdette: not sure you should be bumping this manually

PR #40, opened 2017-09-15
STITCH-539 - Auth methods return only authed userId in body

makesitgo:
* Remove `login( ... )` method from anon & userpass providers (replace with authenticate methods)
* Move "if auth exists, re-use" logic to Stitch client's authenticate method (rather than living in auth provider's login method)
* Returns `json.userId` from Stitch client's authenticate method

### DIFF
--- a/src/auth/providers.js
+++ b/src/auth/providers.js
@@ -27,13 +27,12 @@ function getDeviceInfo(deviceId, appId, appVersion = '') {
 
 function anonProvider(auth) {
   return {
-    login: (opts) => {
-      // reuse existing auth if present
-      const authData = auth.get();
-      if (authData.hasOwnProperty('accessToken')) {
-        return Promise.resolve(authData);
-      }
-
+    /**
+     * Login to a stitch application using anonymous authentication
+     *
+     * @returns {Promise} a promise that resolves when authentication succeeds.
+     */
+    authenticate: () => {
       const device = getDeviceInfo(auth.getDeviceId(), !!auth.client && auth.client.clientAppID);
       const fetchArgs = common.makeFetchArgs('GET');
       fetchArgs.cors = true;
@@ -53,10 +52,11 @@ function userPassProvider(auth) {
      *
      * @param {String} username the username to use for authentication
      * @param {String} password the password to use for authentication
-     * @returns {Promise}
+     * @returns {Promise} a promise that resolves when authentication succeeds.
      */
-    login: (username, password, opts) => {
+    authenticate: ({ username, password }) => {
       const device = getDeviceInfo(auth.getDeviceId(), !!auth.client && auth.client.clientAppID);
+
       const fetchArgs = common.makeFetchArgs(
         'POST',
         JSON.stringify({ username, password, options: { device } })
--- a/src/client.js
+++ b/src/client.js
@@ -68,10 +68,10 @@ class StitchClient {
    */
   login(email, password, options = {}) {
     if (email === undefined || password === undefined) {
-      return this.auth.provider('anon').login(options);
+      return this.authenticate('anon', options);
     }
 
-    return this.auth.provider('userpass').login(email, password, options);
+    return this.authenticate('userpass', Object.assign({ username: email, password }, options));
   }
 
   /**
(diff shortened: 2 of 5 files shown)


makesitgo:
@mpobrien @edaniels ... I took a shot in the dark at this one ... not sure this was the right place to make the change

makesitgo:
cc @mpobrien ... pulling this off from review for a little bit.  discussed a few further changes to make with eric offline

makesitgo:
@mpobrien ... throwing this back up for review if/when you have a chance.  Tests are still failing locally, but I'm not confident it's related to the changes (my local env for testing might be a bit screwy)

test/auth.test.js
```diff
@@ -31,28 +31,36 @@ describe('Auth', () => {
   beforeEach(() => { test.fetch = sinon.stub(window, 'fetch'); });
   afterEach(() => test.fetch.restore());
 
-  it('should allow the user to retrieve profile information', () => {
```
makesitgo: [ ⚠️  ] removed this test

test/test.js
```diff
@@ -220,25 +220,13 @@ describe('Auth', () => {
           .then(() => testClient.logout())
           .then(() => {
             expect(testClient.auth.getDeviceId()).toEqual(mockDeviceId);
-          })
-          .then(() => {
```
makesitgo: [ ⚠️  ] removed this entire block of assertions

test/test.js
```diff
@@ -416,24 +404,6 @@ describe('anonymous auth', () => {
         expect(response.result[0].x).toEqual(1);
       });
   });
-
-  it('includes device info as a query param with anonymous auth method', () => {
```
makesitgo: [ ⚠️  ] removed this entire test

src/auth/providers.js
```diff
@@ -50,13 +44,12 @@ function userPassProvider(auth) {
   return {
     /**
      * Login to a stitch application using username and password authentication
-     *
-     * @param {String} username the username to use for authentication
-     * @param {String} password the password to use for authentication
-     * @returns {Promise}
      */
-    login: (username, password, opts) => {
+    authenticate: data => {
```
mpobrien: Should we write a new JS doc for `authenticate`?

src/client.js
```diff
@@ -58,20 +58,14 @@ class StitchClient {
   }
 
   /**
-   * Login to stich instance, optionally providing a username and password. In
```
mpobrien: Why remove the jsdoc?

src/auth/providers.js
```diff
@@ -50,13 +44,12 @@ function userPassProvider(auth) {
...
+      const {username, password} = data;
```
mpobrien: i think you could just avoid the need for this line by changing L48 to just:
`authenticate : ({username, password}) => ...`

src/client.js
```diff
@@ -58,20 +58,14 @@ class StitchClient {
...
-   * the event that these are omitted, anonymous authentication is used.
-   *
-   * @param {String} [email] the email address used for login
-   * @param {String} [password] the password for the provided email address
-   * @param {Object} [options] additional authentication options
-   * @returns {Promise}
-   */
+  * Convenience method for authenticating with the user/password auth provider
```
mpobrien: This comment isn't entirely accurate - it also falls back to anonymous auth, if either username or password isn't passed.

src/client.js
```diff
@@ -97,8 +91,15 @@ class StitchClient {
    * @param {Object} [options] additional authentication options (user data)
    * @returns {Promise}
    */
-  authenticate(providerType, options) {
```
mpobrien: We need to update the JS doc for this function, a few reasons:
- It's not OAuth specific at all, since you could actually pass in any `providerType`, which we're now doing explicitly in other client functions.
- The additional change in behavior (reusing existing auth if present) is worth mentioning
- A note on the `@returns {Promise}` line that says that the promise resolves to the userId (string type) would be helpful.

edaniels:
Yes, this calls for a major version bump (warranted) and some docs changes. I doubt people were using providers directly but we can still change the docs.

src/client.js
makesitgo: Oversight, originally I had this function entirely removed before realizing the implications of removing a client-level method.  Going to add it back and re-word accordingly.

src/client.js
makesitgo: Oversight, originally I had this function entirely removed before realizing the implications of removing a client-level method. Going to add it back and re-word accordingly.

src/auth/providers.js
makesitgo: After inlining the `({ username, password }) => ...` params, I added back the JS Doc lines describing the params.

src/auth/providers.js
```diff
@@ -27,7 +27,10 @@ function getDeviceInfo(deviceId, appId, appVersion = '') {
 
 function anonProvider(auth) {
   return {
-    authenticate: data => {
+    /**
+     * Login to a stitch application using anonymous authentication
+     */
```
mpobrien: We should als specify that this returns a promise using `@returns` - similar to what's in client.js L67.

src/auth/providers.js
```diff
@@ -44,6 +47,9 @@ function userPassProvider(auth) {
   return {
     /**
      * Login to a stitch application using username and password authentication
+     *
+     * @param {String} username the username to use for authentication
+     * @param {String} password the password to use for authentication
```
mpobrien: Similar to above comment: `@returns {Promise} a promise that resolves when authentication succeeds.

makesitgo:
@mpobrien ... ready for another review.  updated the javascript docs, but please suggest changes to the wording as you see fit.

currently tests are failing, but I believe it is unrelated to my changes (maybe related to the api v2 changes that have been going on?  cc: @edaniels ):

```cmd
[pipeline.go:ExecutePipeline:227] Error executing pipeline: Pipeline stage #0 (mongodb service 'mdb1') failed during execution - Failed to apply update: Document is corrupted
```

mpobrien:
Test issues sorted out offline.
Everything else LGTM here 👍

PR #42, opened 2017-09-27
Allow override of public+client API version subpaths (defaults to 1.0)

mpobrien:
This is a precursor to splitting off the JS SDK's admin api, and allowing it to call against the v2 endpoints

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -30,10 +30,20 @@ class StitchClient {
 
     this.clientAppID = clientAppID;
 
-    this.appUrl = `${baseUrl}/api/public/v1.0`;
-    this.authUrl = `${baseUrl}/api/public/v1.0/auth`;
+    let publicAPIVersionSubpath = 'v1.0';
+    if (options && options.publicAPIVersionSubpath) {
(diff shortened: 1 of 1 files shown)


src/client.js
```diff
@@ -30,10 +30,20 @@ class StitchClient {
...
+      publicAPIVersionSubpath = options.publicAPIVersionSubpath;
+    }
+
+    let clientAPIVersionSubpath = 'v1.0';
+    if (options && options.clientAPIVersionSubpath) {
+      clientAPIVersionSubpath = options.clientAPIVersionSubpath;
+    }
+
+    this.appUrl = `${baseUrl}/api/public/${publicAPIVersionSubpath}`;
+    this.authUrl = `${baseUrl}/api/public/${publicAPIVersionSubpath}/auth`;
```
edaniels: This URL won't be exactly the same in v2. It will be `/auth/<providerName>/login`

mpobrien:
Actually, changed my mind about this approach - this would require keeping around two instances of the client around (one for v1, one for v2) which is confusing and potentially buggy; instead, it should be possible to re-use a single client to access both api versions to make incremental migration simpler, so I'm working on that instead.

PR #43, opened 2017-09-29
STITCH-759 setup v2 structure for JS sdk, implement app and values, w…

mpobrien:
…ith tests

### DIFF
--- a/.evg.yml
+++ b/.evg.yml
@@ -76,7 +76,7 @@ tasks:
             echo "running client tests"
             cd stitch-js-sdk
             npm install
-            npm test
+            npm test -- --runInBand # force serial execution
 
   - name: dist
     depends_on:
--- a/src/client.js
+++ b/src/client.js
@@ -15,6 +15,9 @@ import {
 
 const EJSON = new ExtJSONModule();
 
+const v1 = 1;
+const v2 = 2;
+
 /**
  * Create a new StitchClient instance.
  *
@@ -30,12 +33,28 @@ class StitchClient {
 
     this.clientAppID = clientAppID;
 
-    this.appUrl = `${baseUrl}/api/public/v1.0`;
-    this.authUrl = `${baseUrl}/api/public/v1.0/auth`;
-    if (clientAppID) {
-      this.appUrl = `${baseUrl}/api/client/v1.0/app/${clientAppID}`;
-      this.authUrl = `${this.appUrl}/auth`;
-    }
+    this.authUrl = (
+      clientAppID ?
+        `${baseUrl}/api/client/v1.0/app/${clientAppID}/auth` :
+        `${baseUrl}/api/public/v1.0/auth`
+    );
+
+    this.rootURLsByAPIVersion = {
+      [v1]: {
+        public: `${baseUrl}/api/public/v1.0`,
+        client: `${baseUrl}/api/client/v1.0`,
+        app: (clientAppID ?
+              `${baseUrl}/api/client/v1.0/app/${clientAppID}` :
+              `${baseUrl}/api/public/v1.0`)
+      },
+      [v2]: {
+        public: `${baseUrl}/api/public/v2.0`,
+        client: `${baseUrl}/api/client/v2.0`,
+        app: (clientAppID ?
+              `${baseUrl}/api/client/v2.0/app/${clientAppID}` :
+              `${baseUrl}/api/public/v2.0`)
+      }
+    };
 
     this.auth = new Auth(this, this.authUrl);
     this.auth.handleRedirect();
@@ -221,7 +240,8 @@ class StitchClient {
   _do(resource, method, options) {
     options = Object.assign({}, {
       refreshOnFailure: true,
-      useRefreshToken: false
+      useRefreshToken: false,
+      apiVersion: v1
     }, options);
 
     if (!options.noAuth) {
@@ -230,7 +250,8 @@ class StitchClient {
       }
     }
 
-    let url = `${this.appUrl}${resource}`;
+    const appURL = this.rootURLsByAPIVersion[options.apiVersion].app;
+    let url = `${appURL}${resource}`;
     let fetchArgs = common.makeFetchArgs(method, options.body);
 
     if (!!options.headers) {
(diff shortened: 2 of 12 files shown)


.evg.yml
```diff
@@ -76,7 +76,7 @@ tasks:
             echo "running client tests"
             cd stitch-js-sdk
             npm install
-            npm test
+            npm test -- --runInBand # force serial execution
```
mpobrien: Note: Added this as a workaround for an occasional race condition that causes the mongodb service test to fail (Jest runs all test suites in parallel by default). I spent a bit of time trying to hunt down the root cause to no avail, but for the sake of unblocking this ticket + all the other V2 work, this bandaid should be sufficient, and I'll do some further test cleanup work separately which should take care of it. Running in serial only adds a negligible amount to the time taken to run the tests, since most of the suites complete on the order of a few ms.

mpobrien:
This establishes a structure for the V2 api, along with some refactoring.
After this build runs, I'm also going to try a patch build on the admin code with this version of the JS sdk to verify that it still comes out clean as well.

.evg.yml
edaniels: Sounds good. Maybe file a ticket to come back to this in case anyone is ever testing the SDK outside of MongoDB.

src/client.js
```diff
@@ -30,12 +33,28 @@ class StitchClient {
...
+      [v1]: {
+        public: `${baseUrl}/api/public/v1.0`,
+        client: `${baseUrl}/api/client/v1.0`,
+        app: (clientAppID ?
+              `${baseUrl}/api/client/v1.0/app/${clientAppID}` :
+              `${baseUrl}/api/public/v1.0`)
+      },
+      [v2]: {
+        public: `${baseUrl}/api/public/v2.0`,
+        client: `${baseUrl}/api/client/v2.0`,
```
edaniels: [q] Should we bother including the two client references here when the API doesn't actually exist?

src/client.js
```diff
@@ -58,7 +77,7 @@ class StitchClient {
   }
 
   /**
-   * Login to stich instance, optionally providing a username and password. In
+   * Login to stitch instance, optionally providing a username and password. In
```
edaniels: [lol]

test/admin/testutil.js
```diff
@@ -0,0 +1,7 @@
+const stitch = require('../../src');
+
+export async function getAuthenticatedClient(apiKey) {
+  const adminClient = new stitch.Admin('http://localhost:7080');
```
edaniels: [q] Should this be made configurable?

edaniels: [q] Does there exist a follow up ticket to ensure we add the auth portion of v2 to the SDK?

src/client.js
```diff
@@ -307,38 +328,65 @@ class Admin {
...
-      .then(response => response.json());
-  }
-
-  _get(url, queryParams) {
-    return this._do(url, 'GET', {queryParams});
-  }
-
-  _put(url, options) {
-    return this._do(url, 'PUT', options);
+  get _v2() {
```
edaniels: [q] What's the intent of making these getters? Sugar?

src/client.js
```diff
@@ -487,6 +535,73 @@ class Admin {
...
+                service: (serviceId) => ({
+                  get: () => api._get(`${appUrl}/services/${serviceId}`),
+                  remove: () => api._delete(`${appUrl}/services/${serviceId}`),
+                  config: ()=> ({
+                    get: () => api._get(`${appUrl}/services/${serviceId}/config`),
+                    update: (data) => api._patch(`${appUrl}/services/${serviceId}/config`, data)
+                  })
+                })
+              }),
+              pushNotifications: TODOnotImplemented,
```
edaniels: Nice, I like this

src/client.js
```diff
@@ -487,6 +535,73 @@ class Admin {
...
+              users: TODOnotImplemented,
+              sandbox: TODOnotImplemented,
```
edaniels: I think we are calling this dev in the future, right?

test/admin/service.test.js
```diff
@@ -0,0 +1,74 @@
+const StitchMongoFixture = require('../fixtures/stitch_mongo_fixture');
```
edaniels: [req] Can you add a few failure tests for:
* Invalid service name
* Unknown service type
* Invalid config

test/fixtures/stitch_mongo_fixture.js
```diff
@@ -34,7 +34,9 @@ export default class StitchMongoFixture {
     this.stitch = new StitchService;
   }
 
-  async setup() {
+  async setup(options) {
```
edaniels: This might be out of scope, but the use of this function now seems kind of odd. I'd expect that having createApp be true would create a new empty app, not import one over. Thoughts on this?

test/admin/values.test.js
```diff
@@ -0,0 +1,59 @@
...
+  });
+  it('creating value should make it appear in list', async () => {
+    let value = await appValues.create({name: testValueName});
+    expect(value.name).toEqual(testValueName);
+    let values = await appValues.list();
+    expect(values).toHaveLength(1);
+    expect(values[0]).toEqual(value);
+  });
+  it('fetching a value returns deep value data', async () => {
+    let value = await appValues.create({name: testValueName, value: '"foo"'});
```
edaniels: [q] Should working with values in the SDK be slightly less raw such that you can pass it any value that we will then extend the JSON on? For example, this test would insert `"foo"`

src/client.js
mpobrien: yeah, sugar. this method is not consumed outside the Admin class.

src/client.js
mpobrien: yes, true. updated.

test/admin/testutil.js
mpobrien: it's basically just hardcoded to match what's in the test config file in `test/configs/test_config.json`

edaniels: 👍

test/admin/service.test.js
mpobrien: ✅

test/fixtures/stitch_mongo_fixture.js
mpobrien: I think it's out of scope for this ticket but I definitely agree; I filed a ticket (STITCH-779) to refactor the test fixtures a bit. I would like to get rid of test_app_config.json entirely and rip out the code for starting up the test server, since we don't use it at all.

test/admin/values.test.js
mpobrien: So, this test is a little misleading, because it should just be `'foo'` instead of `'"foo"'` (updated), so it actually does just that - in other words, the body of the request is the result of `JSON.stringify(args)`

src/client.js
mpobrien: Well, there's no way to actually use the client v2 API from the SDK right now anyway, since its methods (executePipeline, for instance) still fall using v1, so I don't think it matters; once the API does exist though, we're going to need to add this anyway, so might as well just include it now.

edaniels: Sounds good

test/admin/values.test.js
edaniels: oh. got it!

PR #46, opened 2017-10-05
STITCH-762 - Adding API V2 endpoints for Auth Providers

haleyowen:


test/admin/authProviders.test.js
```diff
@@ -0,0 +1,97 @@
...
+
+  it('listing auth providers should return empty list', async () => {
+    let providers = await authProviders.list();
+    expect(providers).toEqual([]);
+  });
+  it('creating auth provider should work', async () => {
+    let newProvider = await authProviders.create({type: 'local-userpass', config: validConfig});
+    expect(newProvider.type).toEqual('local-userpass');
+    expect(newProvider.name).toEqual('local-userpass');
+    expect(newProvider.config).toEqual(undefined);
```
mpobrien: nit: use `.toBeUndefined()` here instead.

test/admin/authProviders.test.js
```diff
@@ -0,0 +1,97 @@
...
+  });
+  it('fetching auth provider should work', async () => {
+    let newAuthProvider = await authProviders.create({type: 'local-userpass', config: validConfig});
+    expect(newAuthProvider.type).toEqual('local-userpass');
+    let provider = await authProviders.authProvider(newAuthProvider._id).get();
+    expect(provider.type).toEqual(provider.type);
+  });
+  it('enabling auth provider should work', async () => {
+    let newAuthProvider = await authProviders.create({type: 'local-userpass', config: validConfig, disabled: true});
+    expect(newAuthProvider.type).toEqual('local-userpass');
```
mpobrien: It might be worth doing an additional assertion here to verify that `disabled` is actually `true` for the newly created key, so we know that the flag is actually effective when used as part of the `create()` method, and that the `enable()` call actually had an effect.
Similar for the next test too, after L61. We should check that `disabled` is `false`, before proceeding to call `disable` and verifying its transition to `true`.

PR #50, opened 2017-10-06
STITCH-764: Debug Console Pipeline Execution

UnicodeSnowman:
* Implement `dev.executePipeline` method
  * Update `v2._post` to accept query params in order to support `user_id` needed for dev pipeline execution
* Add simple test to ensure execution works

### TODO
* [x] Blocked by https://github.com/mongodb/stitch-js-sdk/pull/49

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -345,14 +345,17 @@ class Admin {
       _get: (url, queryParams) => v2do(url, 'GET', {queryParams}),
       _put: (url, data) =>
         (data ?
-          v2do(url, 'PUT', {body: JSON.stringify(data)})  :
+          v2do(url, 'PUT', {body: JSON.stringify(data)}) :
           v2do(url, 'PUT')),
       _patch: (url, data) =>
         (data ?
-          v2do(url, 'PATCH', {body: JSON.stringify(data)})  :
+          v2do(url, 'PATCH', {body: JSON.stringify(data)}) :
           v2do(url, 'PATCH')),
       _delete: (url)  => v2do(url, 'DELETE'),
-      _post: (url, body) => v2do(url, 'POST', {body: JSON.stringify(body)})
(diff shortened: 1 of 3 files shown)


src/client.js
```diff
@@ -589,8 +592,19 @@ class Admin {
...
               pushNotifications: TODOnotImplemented,
-              users: TODOnotImplemented,
-              dev: TODOnotImplemented,
+              users: () => ({
+                list: (filter) => api._get(`${appUrl}/users`, filter),
+                create: (user) => api._post(`${appUrl}/users`, user),
+                user: TODOnotImplemented
+              }),
+              dev: () => ({
+                executePipeline: (body, userId, options) => {
```
UnicodeSnowman: on a related note, the v1 version of this method accepted `options`, but I wasn't 100% sure of what those options actually are. Is this needed and/or even valid? cc @mpobrien

mpobrien: It does get used in one additional place: https://github.com/10gen/stitch-ui/blob/105abcbe2965a42ab08c33688a6bbf7936d443af/src/push/actions.js#L25

For the push notifications page, we actually use the sandbox to do the sending of the push notifications. The option specified is just a query param that indicates to the server that it should run the pipeline as a system user and bypass rules.

test/admin/dev.test.js
```diff
@@ -0,0 +1,99 @@
...
+    output: 'array'
+  };
+
+  const pipeline = await appPipelines.create(testPipeline);
+  expect(pipeline.name).toEqual(pipelineName);
+  const pipelines = await appPipelines.list();
+  expect(pipelines).toHaveLength(1);
+}
+
+async function createAppUser(users, { email = "dude.mcgee@doofus.net", password = "doofus123" } = {}) {
```
mpobrien: lol

PR #53, opened 2017-10-09
Fix auth providers crud sdk test

mpobrien:
@haleyowen 
Since we made the server-side change to include disabled auth providers, I think this test needed an update. I updated the change for enable/disable to what I think it should be; it seems to fix the "disable auth provider" test, but not the "enable" one - I'm not sure why. Any ideas?


haleyowen:
Just merged mine, this should be good to go!

PR #55, opened 2017-10-13
STITCH-765 - Adds push notifications to v2 client & tests

makesitgo:
Will fix the typo in main commit once rebase is complete.
Will rebase once extra lint issue fixes are approved.

### DIFF
--- a/test/admin/push.test.js
+++ b/test/admin/push.test.js
@@ -1,0 +1,95 @@
+const StitchMongoFixture = require('../fixtures/stitch_mongo_fixture');
+
+import { getAuthenticatedClient } from '../testutil';
+
+describe('Push Notifications V2', () => {
+  let test = new StitchMongoFixture();
+  let pushNotifications;
+  let app;
+  let apps;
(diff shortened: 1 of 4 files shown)


makesitgo:
cc: @mpobrien for a quick look over the lint fixes I put it (unrelated to push notification changes)

cc: @UnicodeSnowman to review and sync up after to figure out how to deliver v2 push notifications

test/admin/push.test.js
```diff
@@ -0,0 +1,111 @@
...
+  const SENT_TYPE = 'sent';
+  const testNotification = {
+    type: DRAFT_TYPE,
+    message: 'this is a test notification.',
+    label: 'test',
+    topic: 'v2'
+  };
+
+  it('listing draft push notifications should return empty list', async () => {
+    expect.assertions(1);
```
mpobrien: You can actually omit `expect.assertions` in any tests where you're not using any branching that would result in a test assertion to get skipped over (like an if/else, try/catch, or .then/.catch). I don't think we need them in this file.

PR #62, opened 2017-10-20
Fix issue where access token infinitely refreshes

edaniels:


src/auth/index.js
```diff
@@ -32,26 +59,22 @@ export default class Auth {
...
 
-    return this.client.doSessionPost().then(json => this.setAccessToken(json.accessToken));
+    return this.client.doSessionPost().then(json => this.set(json));
   }
 
   pageRootUrl() {
     return [window.location.protocol, '//', window.location.host, window.location.pathname].join('');
   }
 
-  setAccessToken(token) {
```
edaniels: Cleanup

src/auth/index.js
```diff
@@ -32,26 +59,22 @@ export default class Auth {
...
-    this.set(currAuth);
-  }
-
   error() {
     return this._error;
   }
 
   isClient() {
-    return !!this.client && (this.client.type === common.CLIENT_TYPE);
+    if (!this.client) {
```
edaniels: Cleanup

src/auth/index.js
```diff
@@ -142,49 +165,59 @@ export default class Auth {
...
 
-  // Returns true if the access token is expired or is going to expire within 'withinSeconds' seconds,
-  // according to current system time. This threshold is 10 seconds by default to account for latency and clock drift.
-  // Returns false if the access token exists and is not expired nor expiring within 'withinSeconds' seconds.
-  // Returns undefined if the access token doesn't exist, is malformed, or does not have an 'exp' field.
-  isAccessTokenExpired(withinSeconds = 10) {
+  // Returns whether or not the access token is expired or is going to expire within 'withinSeconds'
+  // seconds, according to current system time. Returns false if the token is malformed in any way. 
+  isAccessTokenExpired(withinSeconds = DEFAULT_ACCESS_TOKEN_EXPIRE_WITHIN_SECS) {
     let token = this.getAccessToken();
```
edaniels: Cleanup

src/auth/index.js
```diff
@@ -32,26 +59,22 @@ export default class Auth {
       return this.refreshImpersonation(this.client);
     }
 
-    return this.client.doSessionPost().then(json => this.setAccessToken(json.accessToken));
```
edaniels: This was the bug

src/auth/index.js
```diff
@@ -142,49 +165,59 @@ export default class Auth {
...
+    } catch (e) {
+      return false;
     }
-    return false;
+    
+    return decodedToken.exp && Math.floor(Date.now() / 1000) >= decodedToken.exp - withinSeconds;
   }
 
   getAccessToken() {
-    return this.get()['accessToken'] || this.get()['access_token'];
```
edaniels: This was also a bug

src/auth/index.js
```diff
@@ -7,6 +7,27 @@ import * as common from '../common';
...
+  [REFRESH_TOKEN_FIELD]: 'refreshToken',
+  [DEVICE_ID_FIELD]: 'deviceId',
+  [USER_ID_FIELD]: 'userId'
+};
+
+const adminJSONFields = {
+  [ACCESS_TOKEN_FIELD]: 'access_token',
+  [REFRESH_TOKEN_FIELD]: 'refresh_token',
+  [DEVICE_ID_FIELD]: 'device_id',
+  [USER_ID_FIELD]: 'user_id'
```
mburdette: I don't think these should be in `auth` I think these should be passed in when `auth` is constructed and then that would set `jsonFields` on `auth`.

mburdette: The main issue is really that `Auth` has a `this.client` and `Client` has a `this.auth`

edaniels: I disagree that these should be passed in. Auth itself has ownership of the logic here. It would be poor encapsulation to pass that configuration in.

edaniels: Auth receiving a client is unrelated to this

edaniels:
@mburdette PTAL

src/common.js
```diff
@@ -1,14 +1,4 @@
 export const JSONTYPE = 'application/json';
-export const USER_AUTH_KEY = '_stitch_ua';
```
edaniels: Moved to auth common

src/client.js
```diff
@@ -128,7 +128,7 @@ export default class StitchClient {
     }
 
     return this.auth.provider(providerType).authenticate(options)
-      .then(authData => authData.userId);
+      .then(() => this.auth.authedId());
```
edaniels: Reaching into unmodeled data here

src/admin.js
```diff
@@ -2,7 +2,7 @@
 /* eslint no-labels: ['error', { 'allowLoop': true }] */
 import 'fetch-everywhere';
 import StitchClient from './client';
-import ADMIN_TYPE from './common';
+import ADMIN_CLIENT_TYPE from './common';
```
edaniels: I couldn't resist

src/auth/index.js
```diff
@@ -32,26 +39,22 @@ export default class Auth {
...
-    this.set(currAuth);
-  }
-
   error() {
     return this._error;
   }
 
-  isClient() {
-    return !!this.client && (this.client.type === common.CLIENT_TYPE);
+  isAppClient() {
```
edaniels: I think this name makes it easier to reason about what we are doing

src/auth/index.js
```diff
@@ -121,75 +124,92 @@ export default class Auth {
...
-    return this.get()['accessToken'] || this.get()['access_token'];
+    return this._get().accessToken;
   }
 
   getRefreshToken() {
-    return this.storage.get(common.REFRESH_TOKEN_KEY);
+    return this.storage.get(authCommon.REFRESH_TOKEN_KEY);
   }
 
   set(json) {
```
edaniels: This function is now the *only* place where we use `jsonFields`

edaniels:
@mburdette made some more changes from your feedback. I'd take a fresh look

src/auth/index.js
```diff
@@ -3,18 +3,21 @@
 import { createStorage } from './storage';
 import { createProviders } from './providers';
 import { StitchError } from '../errors';
+import * as authCommon from './common';
 import * as common from '../common';
 
 const jwtDecode = require('jwt-decode');
 
 export default class Auth {
   constructor(client, rootUrl, options) {
```
mburdette: could just use default values in constructor instead of using the `Object.assign` below. You can nest default values. 

`constructor(client, rootUrl, options = {storageType = 'localStorage', codec = authCommon.APP_CLIENT_CODEC})`

this gives a good idea to the consumer what kind of options this can take.

src/client.js
```diff
@@ -56,7 +57,11 @@ export default class StitchClient {
       }
     };
 
-    this.auth = new Auth(this, this.authUrl);
+    const authOptions = {codec: APP_CLIENT_CODEC};
+    if (options && options.authCodec) {
```
mburdette: none of this is needed with default values in the constructor

src/admin.js
```diff
@@ -2,18 +2,19 @@
...
+import { ADMIN_CLIENT_CODEC } from './auth/common';
 
 const v1 = 1;
 const v2 = 2;
 
 export default class Admin extends StitchClient {
   constructor(baseUrl) {
-    super('', {baseUrl});
+    super('', {baseUrl, authCodec: ADMIN_CLIENT_CODEC});
```
mburdette: I don't think this is the way to do this. I think ideally Admin sets up its own `Auth` by doing `this.auth` or you call super with a `authConfig` with contains `url` and `codec`.

src/auth/providers.js
```diff
@@ -55,8 +55,8 @@ function anonProvider(auth) {
 
 /** @namespace */
 function userPassProvider(auth) {
-  const providerRoute = auth.isClient() ? 'local/userpass' : 'providers/local-userpass';
-  const loginRoute = auth.isClient() ? 'local/userpass' : `${providerRoute}/login`;
+  const providerRoute = auth.isAppClient() ? 'local/userpass' : 'providers/local-userpass';
```
mburdette: [nit] `auth.isClientApp()` may be better?

src/client.js
```diff
@@ -56,7 +57,11 @@ export default class StitchClient {
...
+      authOptions.codec = options.authCodec;
+    }
+    this.auth = new Auth(this, this.authUrl, authOptions);
```
mburdette: I think `authUrl` should be in the constructor inside of an `authConfig` object as `url`. Same for options as `options`. we should try to make `StitchClient` not have logic to for admin functionality and let `Admin` handle it.

src/auth/index.js
edaniels: This won't work when only one option is specified. In this case the codec would get unset.

mburdette:
@edaniels a lot of the points here are a bit out of scope, I think it would probably be good to just get the fixes in and then I can go through and make most of the optimizations in one move.

edaniels:
Okay. Merging. I agree that StitchClient contains too much logic for admin.

PR #65, opened 2017-10-26
Fix cloud UA unmarshal

edaniels:
A fix of a similar vein. Cloud login was not working on dev due to us assuming the keys of the model instead of using the codec when doing `unmarshallUserAuth`.


src/auth/index.js
```diff
@@ -274,4 +274,65 @@ export default class Auth {
     this.storage.remove(authCommon.IMPERSONATION_USER_KEY);
     this.storage.remove(authCommon.IMPERSONATION_REAL_USER_AUTH_KEY);
   }
+
+  parseRedirectFragment(fragment, ourState) {
```
edaniels: I don't love this here, but we are breaking dev and therefore breaking any cloud testing on dev (Harry noticed this)

src/auth/index.js
```diff
@@ -274,4 +274,65 @@ export default class Auth {
...
+    // After being redirected from oauth, the URL will look like:
+    // https://todo.examples.stitch.mongodb.com/#_stitch_state=...&_stitch_ua=...
+    // This function parses out stitch-specific tokens from the fragment and
+    // builds an object describing the result.
+    const vars = fragment.split('&');
+    const result = { ua: null, found: false, stateValid: false, lastError: null };
+    let shouldBreak = false;
+    for (let i = 0; i < vars.length; ++i) {
```
mburdette: [nit] you could use `vars.forEach(var =>`

src/auth/index.js
```diff
@@ -274,4 +274,65 @@ export default class Auth {
...
+      const pairParts = vars[i].split('=');
+      const pairKey = decodeURIComponent(pairParts[0]);
+      switch (pairKey) {
+      case authCommon.STITCH_ERROR_KEY:
+        result.lastError = decodeURIComponent(pairParts[1]);
```
mburdette: [semi-nit] I think i would prefer making another const for this `pairPart` just to be more descriptive.

src/auth/index.js
```diff
@@ -274,4 +274,65 @@ export default class Auth {
...
+        result.found = true;
+        let theirState = decodeURIComponent(pairParts[1]);
+        if (ourState && ourState === theirState) {
+          result.stateValid = true;
+        }
+        continue;
+      default: continue;
+      }
+
+      if (shouldBreak) {
```
mburdette: you could just add this to the for loop conditional i think

src/auth/index.js
```diff
@@ -274,4 +274,65 @@ export default class Auth {
...
+        break;
+      }
+    }
+
+    return result;
+  }
+
+  unmarshallUserAuth(data) {
+    let parts = data.split('$');
+    if (parts.length !== 4) {
```
mburdette: [nit] might want to put this `4` in a const somewhere higher up in the file, although not super important

PR #66, opened 2017-10-26
STITCH-826: Adding function execution

haleyowen:


test/test.js
```diff
@@ -728,3 +729,75 @@ describe('pipeline execution', () => {
...
+          fetchMock.post(LOCALAUTH_URL, {userId: hexStr});
+          fetchMock.post(FUNCTION_URL, (name, arg1, arg2) => {
+            requestArg = arg1;
+            return {result: [{x: {'$oid': hexStr}}]};
+          });
+        });
+
+        it('should encode objects to extended json for outgoing function request body', () => {
+          expect.assertions(1);
+          let requestBodyExtJSON = {name: 'testfunc', arguments: [{x: {'$oid': hexStr}}, 'hello']};
```
mpobrien: nit: i think you could move this directly into an arg to `toEqual` on L773, since we don't actually use it anywhere else so its not necessary to assign it with `let`.

test/test.js
```diff
@@ -728,3 +729,75 @@ describe('pipeline execution', () => {
...
+            .then((response) => expect(response.result[0].x).toEqual(new ejson.bson.ObjectID(hexStr)));
+        });
+      });
+
+      describe('extended json encode (outgoing)', () => {
+        let requestArg;
+        beforeAll(() => {
+          fetchMock.restore();
+          fetchMock.post(LOCALAUTH_URL, {userId: hexStr});
+          fetchMock.post(FUNCTION_URL, (name, arg1, arg2) => {
```
mpobrien: it looks like we don't use `name, arg1, arg2` at all in the function. do we need them? i'm assuming the intention was to assert that the server receives args of the same value that matches what was provided to the client. But I think we're already doing that with the assertion on L773.

PR #67, opened 2017-10-30
STITCH-855 execute a function with admin api

mburdette:


test/admin/dev.test.js
```diff
@@ -97,3 +97,55 @@ describe('Dev V2', () => {
     ]);
   });
 });
+
+const FUNC_NAME = 'myFunction';
+const FUNC_SOURCE = 'exports = function(arg1, arg2){ return {sum: arg1 + arg2, userId: context.user.id}}';
+const myFunction = () => ({ name: FUNC_NAME, source: FUNC_SOURCE });
```
mpobrien: nit: doesn't look this needs to be a function - you could just use an object literal

test/admin/dev.test.js
```diff
@@ -97,3 +97,55 @@ describe('Dev V2', () => {
...
+
+  it('Supports executing the function', async () => {
+    const result = await dev.executeFunction(
+      user._id,
+      FUNC_NAME,
+      777,
+      23,
+    );
+
+    expect(result.sum).toEqual({'$numberInt': '800'});
```
mpobrien: nit: you could just combine these into a single `expect` on the entire object, for extra strictness.

PR #73, opened 2017-11-06
STITCH-870 - StitchClient for apps uses API v2.0; Admin v3 uses /admin/v3.0

edaniels:
@UnicodeSnowman see for reference of why certain changes were made: https://github.com/10gen/stitch/pull/352

Basically, removed some v1 stuff but not all of it. Skipping tests for pipelines since the client no longer knows much about them.

### DIFF
--- a/src/auth/providers.js
+++ b/src/auth/providers.js
@@ -46,7 +46,7 @@ function anonProvider(auth) {
       const fetchArgs = common.makeFetchArgs('GET');
       fetchArgs.cors = true;
 
-      return fetch(`${auth.rootUrl}/anon/user?device=${uriEncodeObject(device)}`, fetchArgs)
+      return fetch(`${auth.rootUrl}/providers/anon-user/login?device=${uriEncodeObject(device)}`, fetchArgs)
         .then(common.checkStatus)
         .then(response => response.json())
         .then(json => auth.set(json));
@@ -56,8 +56,8 @@ function anonProvider(auth) {
 
 /** @namespace */
 function userPassProvider(auth) {
-  const providerRoute = auth.isAppClient() ? 'local/userpass' : 'providers/local-userpass';
-  const loginRoute = auth.isAppClient() ? 'local/userpass' : `${providerRoute}/login`;
+  const providerRoute = auth.isAppClient() ? 'providers/local-userpass' : 'providers/local-userpass';
+  const loginRoute = auth.isAppClient() ? `${providerRoute}/login` : `${providerRoute}/login`;
 
   return {
     /**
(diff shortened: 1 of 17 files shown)


src/auth/providers.js
```diff
@@ -56,8 +56,8 @@ function anonProvider(auth) {
 
 /** @namespace */
 function userPassProvider(auth) {
-  const providerRoute = auth.isAppClient() ? 'local/userpass' : 'providers/local-userpass';
-  const loginRoute = auth.isAppClient() ? 'local/userpass' : `${providerRoute}/login`;
+  const providerRoute = auth.isAppClient() ? 'providers/local-userpass' : 'providers/local-userpass';
+  const loginRoute = auth.isAppClient() ? `${providerRoute}/login` : `${providerRoute}/login`;
```
UnicodeSnowman: is it worth adding a comment about the ternary weirdness? I could see this being a little confusing otherwise

test/test.js
```diff
@@ -637,30 +576,32 @@ describe('proactive token refresh', () => {
...
-        expect(response.result[0].x).toEqual(new ejson.bson.ObjectID(hexStr));
+        expect(response.x).toEqual(new ejson.bson.ObjectID(hexStr));
       });
   });
 
   it('returns a rejected promise if trying to execute a pipeline without auth', (done) => {
     let testClient = new StitchClient('testapp');
     testClient.logout()
-      .then(() => testClient.executePipeline([{action: 'literal', args: {items: [{x: {'$oid': hexStr}}]}}]))
+      .then(() => testClient.executeFunction('testfunc', {items: [{x: {'oid': hexStr}}]}, 'hello'))
```
UnicodeSnowman: why `oid` instead of `$oid`?

edaniels: Took this from previous usages of executeFunction. Will change.

edaniels:
@UnicodeSnowman ready for another look.

src/auth/providers.js
```diff
@@ -56,8 +56,10 @@ function anonProvider(auth) {
 
 /** @namespace */
 function userPassProvider(auth) {
-  const providerRoute = auth.isAppClient() ? 'local/userpass' : 'providers/local-userpass';
-  const loginRoute = auth.isAppClient() ? 'local/userpass' : `${providerRoute}/login`;
+  // The ternary expression here is redundant but is just preserving previous behavior based on whether or not
+  // the client is for the admin or client API.
+  const providerRoute = auth.isAppClient() ? 'providers/local-userpass' : 'providers/local-userpass';
```
mburdette: 😆

PR #77, opened 2017-11-14
Fix StitchError stack traces in node

UnicodeSnowman:
* Add `babel-plugin-transform-builtin-extend`
* Subclassing builtin types needs special treatment by babel
  * Without this plugin, in a node environment, a subclass of Error does not actually behave like a true subclass, which manifests itself via a missing stacktrace when a StitchError is thrown

#### More details [here](https://github.com/babel/babel/issues/3083)

We've been seeing a handful of flaky tests in evergreen caused by uncaught promises. Because of the missing stacktraces, these errors are super hard to track down. This should help with that.

### Before
screen shot 2017-11-14 at 3 20 48 pm

### After
screen shot 2017-11-14 at 3 20 11 pm

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -6,11 +6,12 @@
         "node": "6"
       },
       "useBuiltins": "usage",
-    }]
+    }],
   ],
   "plugins": [
     "transform-async-to-generator",
     "babel-plugin-add-module-exports",
-    "version-transform"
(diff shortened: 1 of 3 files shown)


UnicodeSnowman:
test failures are from currently-broken `code-exec` branch

PR #82, opened 2017-11-17
Add debug.executeFunctionSource

edaniels:


src/admin.js
```diff
@@ -224,12 +224,18 @@ export default class Admin extends StitchClient {
...
-          dev: () => ({
+          debug: () => ({
             executeFunction: (userId, name = '', ...args) => {
               return api._post(
-                `${appUrl}/dev/function`,
+                `${appUrl}/debug/execute_function`,
                 {name, 'arguments': args},
                 { user_id: userId });
+            },
+            executeFunctionSource: (userId, source = '', evalSource = '') => {
```
mburdette: [q] this isn't 100% clear to me, so eval source is what comes from the debug console?

edaniels: yes

mburdette: 👍

PR #83, opened 2017-11-17
Create Client API v2.0 for JavaScript SDK

makesitgo:


src/services/mongodb/collection.js
```diff
@@ -143,109 +143,56 @@ Collection.prototype.upsert =
 
 // private
 function insertOp(self, docs, options) {
-  let stages = [];
-
-  // there may be no docs, when building for chained pipeline stages in
-  // which case the source is considered to be the previous stage
-  if (docs) {
```
makesitgo: [ ⚠️  ]: no longer leaving the potential for no `docs` to be provided to this method, as chaining stages is no longer supported

test/admin/dev.test.js
```diff
@@ -39,37 +47,37 @@ describe('Dev Function', () => {
...
-      user._id,
+    const result = await client.executeFunction(
       FUNC_NAME,
       777,
       23,
     );
 
-    expect(result.result.sum).toEqual({'$numberDouble': '800'});
-    expect(result.result.userId).toEqual(user._id);
+    expect(result.sum).toEqual(800);
```
makesitgo: [ ⚠️  ]: calling out this test assertion change

edaniels: 800 because of decoding extended json?

edaniels:
@makesitgo you should rebase off of master and set the PR to be based on that

test/admin/dev.test.js
makesitgo: @edaniels ... I'm not entirely sure why I needed to change this, but simply made it based off what I was seeing come in/out from the Stitch backend.

edaniels: Cool

makesitgo: just to tie off loose ends, this change was _needed_ (read: not actually needed) b/c I inadvertently switched the test from using the admin api to the client api.  code has been reverted to use the proper api in #84

PR #84, opened 2017-11-20
Create Client API v2.0 for JavaScript SDK

makesitgo:


src/util.js
```diff
@@ -65,93 +66,22 @@ function deprecate(fn, msg) {
 }
 
 /**
- * Utility method for converting the rest response from services
- * into composable `thenables`. This allows us to use the same
- * API for calling helper methods (single-stage pipelines) and
- * pipeline building.
+ * Utility method executing a service action as a function call.
```
mpobrien: typo: "Utility method *for* executing a service action..."

src/services/mongodb/collection.js
```diff
@@ -143,109 +143,56 @@ Collection.prototype.upsert =
```
mpobrien: let's take away `upsert`. this is a good opportunity to remove deprecated methods, so we should go for it. Then we won't need to import `deprecate` anymore, either.

mpobrien: It looks like  `getCollection` and `getDB`/`getDb` can get removed too (in `src/services/mongodb/db.js` and `src/services/mongodb/mongodb_service.js`).

src/services/aws/sqs_service.js
```diff
@@ -0,0 +1,34 @@
+import { serviceResponse } from '../../util';
```
mpobrien: Reminder: this file shouldn't be here - sqs_service.js is removed. Should go away with your rebase on master.

src/client.js
```diff
@@ -206,15 +206,32 @@ export default class StitchClient {
...
+    });
+  }
 
-    return this._do('/functions/call', 'POST', { body: responseEncoder(functionJson) })
+  /**
+   * Executes a service.
+   *
+   * @param {String} service The name of the service.
+   * @param {String} action The name of the service action.
+   * @param {Object} [args] Arguments to pass to the service action.
```
mpobrien: According to https://stackoverflow.com/questions/4729516/correct-way-to-document-open-ended-argument-functions-in-jsdoc the right way to document the variadic args would look like this:

`@param {...*} args Arguments to pass to the service action.`

src/services/mongodb/collection.js
```diff
@@ -143,109 +143,56 @@ Collection.prototype.upsert =
 
 // private
 function insertOp(self, docs, options) {
```
mpobrien: we already discussed offline, but summary of changes to make here for the new crud stuff:

- `insert` is now split into `insertOne` and `insertMany`. 
    - the payload for `insert` should look like `{database: "db", collection: "collection", document: { ... doctoinsert}}`
    - the payload for `insertMany` should look like `{database: "db", collection: "collection", documents: [doc1, doc2, doc3]}`. However note that `insertMany` is *not* variadic, its argument should be an array.

- the results returned by each CRUD action are as described here: https://jira.mongodb.org/browse/STITCH-902. This should be useful in the integration tests.

src/services/mongodb/collection.js
makesitgo: FYI: there was an unused `insert` function (that sat alongside `upsert`) that I removed as well (it just wasn't being deprecated for whatever reason)

edaniels:
Reminder for `errorCode` -> `error_code`

dist/node/client.js
```diff
@@ -271,19 +271,44 @@ var StitchClient = function () {
         args[_key - 1] = arguments[_key];
       }
 
+      return this._doFunctionCall({
```
edaniels: Can you remove the dist from this PR so we can get a minimal view of what work was done? We will update the dist later.

src/client.js
```diff
@@ -203,18 +203,35 @@ export default class StitchClient {
...
+    return this._doFunctionCall({
       name,
       arguments: args
-    };
+    });
+  }
 
-    return this._do('/functions/call', 'POST', { body: responseEncoder(functionJson) })
+  /**
+   * Executes a service.
```
edaniels: [nit] Executes a service function

yarn.lock
```diff
@@ -0,0 +1,4766 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
edaniels: [minor] I didn't think we were switching to yarn in the SDK yet?

src/client.js
```diff
@@ -224,7 +241,6 @@ export default class StitchClient {
    *
    * @returns {Promise}
    */
-
   doSessionPost() {
```
edaniels: I cannot comment on this since it's not in the change, but I don't think `userProfile` works anymore since the user profile is now at `auth/profile`.

yarn.lock
makesitgo: I've been using yarn everywhere so far?  Happy to take this out though.

src/client.js
makesitgo: Is there a good way you want me to test this?  I can't find a test around this function anywhere...  Or should I just /shrug and delete it?

edaniels: Well we can't delete it. It's a needed method to get your profile. This is suitable as an integration test that you run after login. The profile contains details about your identities and such.

src/services/mongodb/collection.js
```diff
@@ -112,140 +108,97 @@ class Collection {
...
-      return doc;
-    });
-
-    stages.push({
-      action: 'literal',
-      args: {
-        items: docs
-      }
-    });
+  const outgoingOptions = { documents: docs.map(setObjectId) };
```
mpobrien: we don't actually need to do this anymore - the server now assigns object id's to the docs being inserted - can remove.

src/services/mongodb/collection.js
```diff
@@ -112,140 +108,97 @@ class Collection {
...
-   * @param {Number} [options.limit=null] The maximum number of documents to return.
-   * @return {Number} An array of documents.
-   */
-  count(query, options = {}) {
-    return findOp(this, query, Object.assign({}, options, {
-      count: true
-    }), result => !!result.result ? result.result[0] : 0);
-  }
+// private
+function setObjectId(doc) {
```
mpobrien: can remove this now.

src/services/mongodb/collection.js
```diff
@@ -19,47 +19,43 @@ class Collection {
...
    *
    * @method
    * @param {Object} query The query used to match the documents to delete.
-   * @param {Object} [options] Additional options object.
    * @return {Promise<Object, Error>} Returns a Promise for the operation.
    */
-  deleteMany(query, options = {}) {
-    return deleteOp(this, query, Object.assign({}, options, { singleDoc: false }));
+  deleteMany(query) {
+    return deleteOp(this, true, query);
```
mpobrien: i don't know if we get much out of having `deleteOp` as a helper anymore (it's just a one-liner now).
I think you could remove it and just call `return buildResponse("deleteMany", this, buildArgs(service, { query })))` directly. Same for `deleteOne` and `insertOne`/`insertMany` too.

src/services/pubnub/pubnub_service.js
```diff
@@ -0,0 +1,31 @@
+import { serviceResponse } from '../../util';
```
mpobrien: we got rid of pubnub - this file should be gone now

src/services/slack/slack_service.js
```diff
@@ -0,0 +1,42 @@
+import { serviceResponse } from '../../util';
```
mpobrien: this should be gone since slack was deprecated. don't forget to rebase.

src/services/mongodb/collection.js
```diff
@@ -112,140 +108,97 @@ class Collection {
...
+  this.limit = limit;
+  return this;
+};
+
+MongoQuery.prototype.sort = function(sort) {
+  this.sort = sort;
+  return this;
+};
+
+MongoQuery.prototype.then = function(resolve) {
```
mpobrien: 👌 slick

src/services/mongodb/collection.js
makesitgo: @mpobrien ... is that because mongodb handles missing object ids?  Just want to make sure I understand why it's not needed anymore...

src/services/pubnub/pubnub_service.js
makesitgo: ugh, i definitely messed up my rebase somewhere along the lines.  same thing happened with sqs, and i just ended up deleting manually.

the following will be removed:
* sqs
* pubnub
* slack

src/services/mongodb/collection.js
mpobrien: yup, the stitch server is populating them now.

mpobrien:
what's the complete case race condition thing about?
otherwise LGTM; nicely done 😊

mpobrien:
One last note - i just merged in support for `count` as a top-level CRUD action at the collection level (here https://github.com/10gen/stitch/commit/b80674089511c99e1ade06859237dba2b2e47582).
We should add that here too along with a couple simple tests (it's just another slight variation on `find`/`findOne`). 
I'm fine with merging what we have here so far and just adding the `count` method in another follow-up commit if you'd prefer; it should be very easy since you already have all the test stuff in place here.

makesitgo:
@mpobrien ... working on the count implementation right now (trying to reuse the `MongoQuery` we defined yesterday.  Should nearly be there.  As for that race condition, I should have updated that comment but before I realized I could use `await` on the `MongoQuery`, I was doing assertions inside a `.then( ... )` block which I think was failing when I'd run many of those tests (it worked when only one was run at a time).  Anyways, that's a non-issue now.

src/util.js
```diff
@@ -6,6 +6,7 @@ const RESULT_METADATA_KEY = '_stitch_metadata';
 /** @namespace util */
 
 /**
+ * TODO: maybe remove me?
```
mpobrien: remove this comment before committing

test/client/mongodb_service.test.js
```diff
@@ -0,0 +1,297 @@
...
+  it('Should be successful for count (with query, with limit)', async() => {
+    const testDocs = [{ a: 1, d: 0 }, { b: 2, d: 0 }, { c: 3, d: 0 }];
+
+    await service.insertMany(testDocs);
+
+    const count = await service.count({ d: 0 }, { limit: 2 });
+    expect(count).toBe(2);
+  });
+
+  // db asertions
```
mpobrien: typo: assertions 😉

PR #88, opened 2017-11-22
fix session + profile root urls

mpobrien:
also some minor refactoring to constify the api types and added a test.

I think ultimately we should work on converting the rest of the test suites in the repo to just use a real server, it'll be less brittle. (see https://jira.mongodb.org/browse/STITCH-922)


edaniels:
LGTM. Please publish as minor

PR #90, opened 2017-11-30
STITCH-931 - HTTP Service Integration Tests

makesitgo:
**Test Output:**
```cmd
 PASS  test/client/http_service.test.js
  Executing http service functions
    That have no matching service rules
      ✓ should fail due to no matching rule found (137ms)
    That have proper service rules created
      And an unavailable mock server
        ✓ should fail due to a function execution error (120ms)
      And an available mock server
        Submitting an invalid request
          ✓ should fail when an invalid url value is supplied (96ms)
          ✓ should fail when an invalid url datatype is supplied (80ms)
          ✓ should fail when both url and scheme+host are supplied (82ms)
        Submitting a GET request
          ✓ with a string argument should send a successful request (114ms)
          ✓ with an object argument should send a successful request (87ms)
        Submitting a POST request
          ✓ with a string argument should send a successful request and return an empty response body (99ms)
          ✓ with an object argument containing the raw request body should send a successful request (110ms)
          ✓ with an object argument containing the stringified request body should send a successful request (87ms)
        Submitting a PUT request
          ✓ with a string argument should send a successful request and return an empty response body (98ms)
          ✓ with an object argument containing the raw request body should send a successful request (83ms)
          ✓ with an object argument containing the stringified request body should send a successful request (85ms)
        Submitting a PATCH request
          ✓ with a string argument should send a successful request and return an empty response body (94ms)
          ✓ with an object argument containing the raw request body should send a successful request (86ms)
          ✓ with an object argument containing the stringified request body should send a successful request (101ms)
        Submitting a DELETE request
          ✓ with a string argument should send a successful request (79ms)
          ✓ with an object argument should send a successful request (83ms)
        Submitting a HEAD request
          ✓ with a string argument should send a successful request (91ms)
          ✓ with an object argument should send a successful request (80ms)
```

### DIFF
--- a/test/client/http_service.test.js
+++ b/test/client/http_service.test.js
@@ -1,0 +1,314 @@
+const SimpleHttpServer = require('../fixtures/simple_http_server');
+const StitchMongoFixture = require('../fixtures/stitch_mongo_fixture');
+
+import { buildClientTestHarness, extractTestFixtureDataPoints } from '../testutil';
+
+describe('Executing http service functions', () => {
+  const server = new SimpleHttpServer();
+  const test = new StitchMongoFixture();
+
+  beforeAll(() => test.setup());
+  afterAll(() => test.teardown());
+
+  const SERVICE_TYPE = 'http';
+  const SERVICE_NAME = 'gd2';
+
+  let th;
+  let service;
+  let serviceId;
+
+  beforeEach(async() => {
+    const { apiKey, groupId, serverUrl } = extractTestFixtureDataPoints(test);
+    th = await buildClientTestHarness(apiKey, groupId, serverUrl);
+
+    const httpService = await th
+      .app()
+      .services()
+      .create({ type: SERVICE_TYPE, name: SERVICE_NAME });
+
+    service = th.stitchClient.service(SERVICE_TYPE, SERVICE_NAME);
+    serviceId = httpService._id;
+  });
+
+  afterEach(() => th.cleanup());
+
+  describe('That have no matching service rules', () => {
+    it('should fail due to no matching rule found', async() => {
+      expect(service.get(server.url)).rejects.toMatchObject({
+        name: 'StitchError',
+        code: 'NoMatchingRuleFound',
+        response: { status: 403 }
+      });
+    });
+  });
+
+  describe('That have proper service rules created', () => {
+    beforeEach(async() => {
+      await th
+        .app()
+        .services()
+        .service(serviceId)
+        .rules()
+        .create({
+          name: 'testRule',
+          actions: ['get', 'post', 'put', 'patch', 'delete', 'head'],
+          when: {}
+        });
+    });
+
+    describe('And an unavailable mock server', () => {
+      it('should fail due to a function execution error', async() => {
+        expect(service.get(server.url)).rejects.toMatchObject({
+          name: 'StitchError',
+          code: 'FunctionExecutionError',
+          response: { status: 400 }
+        });
+      });
+    });
+
+    describe('And an available mock server', () => {
+      beforeAll(() => server.listen());
+      afterAll(() => server.close());
+
+      describe('Submitting an invalid request', () => {
+        it('should fail when an invalid url value is supplied', async() => {
+          expect(service.get({ url: 'invalidurl' })).rejects.toMatchObject({
+            name: 'StitchError',
+            code: 'FunctionExecutionError',
+            response: { status: 400 }
+          });
+        });
+
+        it('should fail when an invalid url datatype is supplied', async() => {
+          expect(service.get({ url: 10281995 })).rejects.toMatchObject({
+            name: 'StitchError',
+            code: 'InvalidParameter',
+            response: { status: 400 }
+          });
+        });
+
+        it('should fail when both url and scheme+host are supplied', async() => {
+          expect(
+            service.get({
+              url: 'http://locahost',
+              scheme: 'http',
+              host: 'localhost'
+            })
+          ).rejects.toMatchObject({
+            name: 'StitchError',
+            code: 'InvalidParameter',
+            response: { status: 400 }
+          });
+        });
+      });
+
+      describe('Submitting a GET request', () => {
+        beforeAll(() => server.mockRequestHandler()
+          .setStatusCode(200)
+          .addHeader('Content-Type', 'text/plain')
+          .addHeader('Warning', '622 Test GET warning')
+          .textResponse('POW!'));
+
+        it('with a string argument should send a successful request', async() => {
+          const { statusCode, headers, body } = await service.get(server.url);
+
+          expect(statusCode).toBe(200);
+          expect(headers['Content-Type']).toContain('text/plain');
+          expect(headers['Warning']).toContain('622 Test GET warning');
+          expect(body.toString()).toBe('POW!');
+        });
+
+        it('with an object argument should send a successful request', async() => {
+          const { statusCode, headers, body } = await service.get({ url: server.url });
+
+          expect(statusCode).toBe(200);
+          expect(headers['Content-Type']).toContain('text/plain');
+          expect(headers['Warning']).toContain('622 Test GET warning');
+          expect(body.toString()).toBe('POW!');
+        });
+      });
+
+      describe('Submitting a POST request', () => {
+        beforeAll(() => server.mockRequestHandler()
+          .setStatusCode(201)
+          .addHeader('Content-Type', 'text/plain')
+          .addHeader('Warning', '622 Test POST warning')
+          .echoResponse());
+
+        it('with a string argument should send a successful request and return an empty response body', async() => {
+          const { statusCode, headers, body } = await service.post(server.url);
+
+          expect(statusCode).toBe(201);
+          expect(headers['Content-Type']).toContain('text/plain');
+          expect(headers['Warning']).toContain('622 Test POST warning');
+          expect(body.toString()).toBe('');
+        });
+
+        it('with an object argument containing the raw request body should send a successful request', async() => {
+          const requestBody = { foo: 'bar' };
+          const { statusCode, headers, body } = await service.post({
+            url: server.url,
+            body: requestBody,
+            encodeBodyAsJSON: true
+          });
+
+          expect(statusCode).toBe(201);
+          expect(headers['Content-Type']).toContain('text/plain');
+          expect(headers['Warning']).toContain('622 Test POST warning');
+          expect(body.toString()).toBe(JSON.stringify(requestBody));
+        });
+
+        it('with an object argument containing the stringified request body should send a successful request', async() => {
+          const requestBody = JSON.stringify({ foo: 'bar' });
+          const { statusCode, headers, body } = await service.post({
+            url: server.url,
+            body: requestBody
+          });
+
+          expect(statusCode).toBe(201);
+          expect(headers['Content-Type']).toContain('text/plain');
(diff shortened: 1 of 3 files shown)


test/client/http_service.test.js
```diff
@@ -0,0 +1,265 @@
...
+    describe('Submitting a GET request', () => {
+      const requestHandler = (req, res) => {
+        res.writeHead(200, { 'Content-Type': 'text/plan' });
+        res.end('POW!');
+      };
+
+      it('with url should work', async() => {
+        handleRequest = requestHandler;
+
+        const { statusCode, body } = await service.get(TEST_SERVER_URL);
```
mpobrien: would be good to include some coverage for the response headers too. stitch will include that data in the service response, so we could set headers in the test server and check them in the result here.

test/client/http_service.test.js
```diff
@@ -0,0 +1,265 @@
...
+          url: TEST_SERVER_URL,
+          body: JSON.stringify({ foo: 'bar' })
+        });
+
+        expect(statusCode).toBe(204);
+        expect(body).toBeFalsy();
+      });
+    });
+
+    describe.skip('Submitting a PATCH request', () => {
```
mpobrien: should this `.skip` be removed?

test/client/http_service.test.js
```diff
@@ -0,0 +1,265 @@
...
+      it('with options should work (stringified body)', async() => {
+        handleRequest = requestHandler;
+
+        const { statusCode, body } = await service.put({
+          url: TEST_SERVER_URL,
+          body: JSON.stringify({ foo: 'bar' })
+        });
+
+        expect(statusCode).toBe(204);
+        expect(body).toBeFalsy();
```
mpobrien: should't we be doing more assertions in this test? the test server doesn't write back anything, so we haven't actually verified anything about if our outgoing data was correct or not (which i think is the intended purpose of this test, based on the description). Maybe the test server should echo the body back to the client, so that you can compare them here, which is what other tests are doing?

test/client/http_service.test.js
```diff
@@ -0,0 +1,265 @@
...
+        handleRequest = requestHandler;
+
+        const { statusCode, body } = await service.delete({ url: TEST_SERVER_URL });
+
+        expect(statusCode).toBe(202);
+        expect(body.toString()).toBe('');
+      });
+    });
+
+    describe.skip('Submitting a HEAD request', () => {
```
mpobrien: is the `.skip` meant to be here?

test/client/http_service.test.js
```diff
@@ -0,0 +1,265 @@
...
+      it('with options should work (stringified body)', async() => {
+        handleRequest = requestHandler;
+
+        const { statusCode, body } = await service.patch({
+          url: TEST_SERVER_URL,
+          body: JSON.stringify({ foo: 'bar' })
+        });
+
+        expect(statusCode).toBe(206);
+        expect(body).toBeFalsy();
```
mpobrien: same note as other (similar) comment: this test doesn't seem to be enforcing anything useful about `body` ?

test/client/http_service.test.js
```diff
@@ -0,0 +1,265 @@
+const http = require('http');
+const StitchMongoFixture = require('../fixtures/stitch_mongo_fixture');
+
+import { buildClientTestHarness, extractTestFixtureDataPoints } from '../testutil';
+
+describe('With a mock server', () => {
```
mpobrien: a test that handles the request when the test server is *not* running would be useful, too.

test/client/http_service.test.js
```diff
@@ -0,0 +1,265 @@
...
+    });
+
+    describe('Submitting a PUT request', () => {
+      const requestHandler = async(req, res) => {
+        const body = await readRequestBody(req);
+        res.writeHead(204, { 'Content-Type': 'text/plan' });
+        res.end(body);
+      };
+
+      it('with url should work', async() => {
```
mpobrien: I think we should make these test descriptions more precise, where possible. For instance, this one is specifically testing that you can invoke `.put` using a string argument instead of an object. So maybe a better description might be "with a string argument should send a successful request", etc.

test/client/http_service.test.js
```diff
@@ -0,0 +1,265 @@
...
+describe('With a mock server', () => {
+  const TEST_SERVER_URL = 'http://localhost:7000';
+
+  let handleRequest;
+  const server = http.createServer((req, res) => handleRequest(req, res));
+
+  beforeAll(() => server.listen(7000));
+  afterAll(() => server.close());
+
+  describe('And http service functions', () => {
```
mpobrien: let's test a few more failure cases, also.
- no matching rules
- bad args (invalid URL for instance, or missing a required option, etc.)

these should trigger `.catch()` on the SDK request.

test/client/http_service.test.js
```diff
@@ -0,0 +1,265 @@
...
+        handleRequest = requestHandler;
+
+        const { statusCode, body } = await service.put(TEST_SERVER_URL);
+
+        expect(statusCode).toBe(204);
+        expect(body).toBeFalsy();
+      });
+
+      it('with options should work (raw body)', async() => {
+        handleRequest = requestHandler;
```
mpobrien: do we need to do `handleRequest = requestHandler` in every function? can we just set it once in a `beforeAll` inside the `describe` block since we're reusing the same one?

test/client/http_service.test.js
```diff
@@ -0,0 +1,265 @@
...
+          body: JSON.stringify({ foo: 'bar' })
+        });
+
+        expect(statusCode).toBe(206);
+        expect(body).toBeFalsy();
+      });
+    });
+
+    describe('Submitting a DELETE request', () => {
+      const requestHandler = (req, res) => {
```
mpobrien: nit/idea: almost all these request handlers are identical or slight variations on the same pattern (set status code + headers + body, done). Maybe we could DRY this up with a helper that just generates these handlers.

test/client/http_service.test.js
```diff
@@ -0,0 +1,265 @@
...
+});
+
+const readRequestBody = req =>
+  new Promise((resolve, reject) => {
+    const body = [];
+    req
+      .on('data', chunk => {
+        body.push(chunk);
+      })
+      .on('end', () => {
```
mpobrien: nit: for completeness here i think you might also want a `.on('error', err => reject(err))`

test/client/http_service.test.js
makesitgo: ah the `beforeAll` was what I was missing... :doh: I was trying to just explicitly set it inside each `describe` block plainly, and ended up stepping on my feet as describes would inevitably end up using other describes' mock handlers.  thanks!

test/client/http_service.test.js
makesitgo: this is related to what i brought up in the sync up: Stitch backend doesn't support `PATCH` or `HEAD` http requests at this point.  would defer to you as to how to handle this in the SDK:

* leave as is (with methods on the http service that are unsupported and skipped tests)
* remove the unsupported methods from the http service (and their skipped tests)
* implement the methods on Stitch's backend (and fill out the tests)

test/client/http_service.test.js
makesitgo: yeah, refer to the comment [above](https://github.com/mongodb/stitch-js-sdk/pull/90#discussion_r154365661)

test/client/http_service.test.js
```diff
@@ -0,0 +1,265 @@
+const http = require('http');
+const StitchMongoFixture = require('../fixtures/stitch_mongo_fixture');
+
+import { buildClientTestHarness, extractTestFixtureDataPoints } from '../testutil';
+
+describe('With a mock server', () => {
+  const TEST_SERVER_URL = 'http://localhost:7000';
+
+  let handleRequest;
```
mpobrien: @makesitgo i have one more idea to offer that might make this set of tests a little bit nicer, but i'll leave it up to you if you think it's worthwhile.
basically, we'd make a small `class` wrapper for the mock server. The class would encapsulate a `requestHandler` var that can be set using a `setRequestHandler` function. Then in your test setup code, you would just call `setRequestHandler()` with the test-specific handler function, instead of assigning `handleRequest` in each block. This makes it a lot clearer that the handler is being injected into the server for test setup purposes. The other benefit of this is that the handler function lives in the same scope as the mock server itself, and we aren't scattering the test setup state across the scope of the entire `describe` block.

test/client/http_service.test.js
mpobrien: Got it. Based on our sync meeting convo it sounds like we ultimately decided to implement PATCH but *not* HEAD. So let's:
- Proceed with what we have here for `patch()` (with the `.skip()` left in)
- File a ticket to add support for PATCH on the server, and include a note in the description about enabling tests for it in the JS sdk (removing `.skip()`) 
- Remove the tests/code for `head()` entirely

edaniels: We're do I just PATCH and HEAD but not OPTIONS. Ticket filed yesterday and at top of queue. My fault for having them removed initially.

test/client/http_service.test.js
makesitgo: I like this ... I think can roll with it

makesitgo:
@mpobrien ... ready for another pass.  I'm keeping PATCH and HEAD tests in there (as skipped for now) while the backend PR is still open

test/fixtures/mock_server.js
```diff
@@ -0,0 +1,77 @@
+const http = require('http');
+
+export default class MockServer {
```
mpobrien: ocd nit: can we name this something like `DummyServer`? technically it's not mocking, it *is* a server.

test/fixtures/mock_server.js
```diff
@@ -0,0 +1,77 @@
...
+
+  close() {
+    this.server.close();
+  }
+
+  listen() {
+    this.server.listen(this.port);
+  }
+
+  url() {
```
mpobrien: optional nit: could make this a getter

test/fixtures/mock_server.js
```diff
@@ -0,0 +1,77 @@
...
+    return `http://localhost:${this.port}`;
+  }
+
+  mockRequestHandler(statusCode, headers, dataCreator) {
+    this.requestHandler = new MockRequestHandler({ statusCode, headers, dataCreator });
+    return this.requestHandler;
+  }
+}
+
+function MockRequestHandler({ statusCode = 500, headers = {}, dataCreator = req => {} }) {
```
mpobrien: cool

test/client/http_service.test.js
makesitgo: Just to tie things up here, I was able to hit Martin's latest changes to support `PATCH` and `HEAD` requests, so these are now good to go

test/client/http_service.test.js
```diff
@@ -0,0 +1,312 @@
...
+    const httpService = await th
+      .app()
+      .services()
+      .create({ type: SERVICE_TYPE, name: SERVICE_NAME });
+
+    service = th.stitchClient.service(SERVICE_TYPE, SERVICE_NAME);
+    serviceId = httpService._id;
+  });
+
+  afterEach(async() => th.cleanup());
```
mpobrien: nit/note: you actually don't need `async` here, since you're not relying on using `await` anywhere in the function body. Plus since the function returns a promise, jest will wait for it to resolve anyway

mpobrien:
Also: Martin's backend PR is merged in, so you can go ahead and take out the skips

mpobrien:
LGTM

PR #91, opened 2017-12-05
STITCH-914 Add custom auth to JavaScript SDK

jsflax:


src/auth/providers.js
```diff
@@ -54,6 +54,36 @@ function anonProvider(auth) {
...
 }
 
+function customProvider(auth) {
+  const providerRoute = 'providers/custom-token';
+  const loginRoute = `${providerRoute}/login`;
+  return {
+    /**
+     * Login to a stitch application using username and password authentication
+     *
+     * @memberof userPassProvider
```
edaniels: This docstring is incorrect

src/auth/providers.js
```diff
@@ -54,6 +54,36 @@ function anonProvider(auth) {
   };
 }
 
+function customProvider(auth) {
+  const providerRoute = 'providers/custom-token';
+  const loginRoute = `${providerRoute}/login`;
+  return {
```
edaniels: newline before this line

src/auth/providers.js
```diff
@@ -331,7 +361,8 @@ function createProviders(auth, options = {}) {
     google: googleProvider(auth),
     facebook: facebookProvider(auth),
     mongodbCloud: mongodbCloudProvider(auth),
-    userpass: userPassProvider(auth)
+    userpass: userPassProvider(auth),
+    custom: customProvider(auth)
```
edaniels: [q] Maybe this should be called customAuth or is that too repetitive?

test/test.js
```diff
@@ -363,6 +365,41 @@ describe('anonymous auth', () => {
...
+
+    fetchMock.post(FUNCTION_CALL_URL, () => {
+      return JSON.stringify({x: ++count});
+    });
+  });
+
+  it('can authenticate with custom auth method', () => {
+    expect.assertions(3);
+    let testClient = new StitchClient('testapp');
+    return testClient.authenticate('custom', 'jwt')
```
edaniels: We should at least verify that the mocked endpoint receives `jwt`

src/auth/providers.js
jsflax: Seems redundant and goes against the existing paradigm of not using `Auth` in the name

src/auth/providers.js
jsflax: Ooops

mburdette: [nit] i think snake casing `userpass` seems weird

edaniels:
bump @jsflax

jsflax:
@edaniels PTAL

PR #94, opened 2017-12-12
STITCH-1007 - Correctly require token as a parameter to custom authentication (fixes #92)

### LINKED ISSUE #92: Authentication
Can't find demo code on authenticating to Stitch using custom authentication.

edaniels:


test/client/http_service.test.js
```diff
@@ -67,8 +67,8 @@ describe('Executing http service functions', () => {
     });
 
     describe('And an available mock server', () => {
-      beforeAll(() => server.listen());
-      afterAll(() => server.close());
+      beforeEach(() => server.listen());
```
edaniels: @makesitgo, using beforeAll here causes the above test where no server is listening to actually pass but have an uncaught exception thrown

PR #96, opened 2017-12-18
STITCH-758 Improve auth in react-native

jsflax:
This ticket is two parts: adding support for flow, to make life easier, and adding support for react native.

To support React Native:
- `StitchClient` now takes an optional `Storage` parameter so that consumers can supply their own storage interface
- `Storage` is now asynchronous. This is actually ideal as most `io` operations should be asynchronous.
- This alters a lot of code internally– in our codebase, `auth` is exposed and used throughout. Externally, the only breaking change is that fetching `authedId` is now an asynchronous operation as the client now needs to read from the local store.
- `TODO`: There will be a separate PR for the actual "React Native SDK" that implements the `Storage` interface that will be filed shortly.
- `TODO`: `Storage` needs to be exposed as an interface.

To support Flow (incomplete):
- Types have been added to any file with `// @flow` added to the top.
- `CodingKeys` have been added to allow a simpler decoding and encoding of JSON so that we can encapsulate our responses better.
- `Provider`s have been given proper types to allow for less loose code.
- `TODO`: Encoding function for `CodingKeys` has not been written, though we have no where to use it yet.
- `TODO`: Add encapsulating classes for the Admin types (Value, App, etc.).


edaniels:
@jsflax can you edit the PR description to give a rundown of what kinds of changes were made to make react native work as well as what we're doing with flow here?

edaniels:
Updating ticket to indicate a major version bump is needed currently from these changes

jsflax:
Updated.

jsflax:
Additionally, there will be some cajoling of the `package.json` to get eslint working alongside flow. Will add to this PR, but should not affect the review.

mburdette:
@jsflax can we not do this in one PR, it should be one PR for the react native stuff, and a different one for the flow stuff.

mpobrien:
Small additional request for when you open new PRs: can we omit the changes in `dist/`? It makes the github UI mostly unusable for code reviewing:(
The js files in `dist/` are all generated at build time anyway, so there's nothing relevant to review.

PR #97, opened 2017-12-18
STITCH-758 Improve Auth in React Native

jsflax:
This forces the Storage interface to be asynchronous.

The breaking change for consumers of this library is that `authedId` will now be an asynchronous call.


src/auth/providers.js
```diff
@@ -289,7 +294,22 @@ function googleProvider(auth) {
      * @param {Object} data the redirectUrl data to use for authentication
      * @returns {Promise} a promise that resolves when authentication succeeds.
      */
-    authenticate: data => {
+    authenticate: async data => {
+      let { authCode } = data;
+      if (authCode !== null) {
```
UnicodeSnowman: FYI, if `data` is equal to `{ authCode: undefined }` or even just `{}`, this check will be truthy. i.e. `undefined !== null`. might be worth defining something like https://github.com/10gen/stitch-ui/blob/master/src/util.js#L176-L184 in the sdk

src/client.js
```diff
@@ -163,7 +173,7 @@ export default class StitchClient {
         useRefreshToken: true,
         rootURL: this.rootURLsByAPIVersion[v2][API_TYPE_CLIENT]
       }
-    ).then(() => this.auth.clear());
+    ).then(async () => await this.auth.clear()); // eslint-disable-line space-before-function-paren
```
mpobrien: i don't think this line needs to change, right?
I think using async/await when you're already inside a promise chain `then()` is probably not optimal, since that will get transpiled into a whole new promise chain.

src/client.js
```diff
@@ -139,22 +149,22 @@ export default class StitchClient {
...
     // reuse existing auth if present
-    if (this.auth.getAccessToken()) {
-      return Promise.resolve(this.auth.authedId());
+    if (await this.auth.getAccessToken()) {
+      return await this.auth.authedId();
     }
 
     return this.auth.provider(providerType).authenticate(options)
-      .then(() => this.auth.authedId());
+      .then(async() => this.auth.authedId());
```
mpobrien: don't think we need async here? should just continue the promise chain.

UnicodeSnowman:
so overall, just another FYI that we should discuss: https://caniuse.com/#feat=async-functions

IE doesn't support async/await, so this will have be polyfilled. this should be handled automagically by `babel-preset-env`, *but* we should make sure we're okay with this, considering that the regenerator runtime is an additional dependency that will now be included in **all** SDK builds, even for those users with browsers that do support generators.

https://babeljs.io/docs/plugins/transform-regenerator/

mpobrien:
Question out of curiosity - what's the change in the bundle sizes with the usage of async/await and the new babel presets? The code generated by the polyfill can in some cases be a bit crazy, so we should make sure we're not inflating the output code by a gross amount.

jsflax:
It's an increase of (123KB -> 163KB) 40KB in the `node` dir, and (2.3MB -> 2.4MB) for the `web` dir.

src/auth/index.js
```diff
@@ -129,149 +131,177 @@ export default class Auth {
...
+      return this.set(realUserAuth);
+    }).then(() => this.clearImpersonation());
   }
 
-  async clearImpersonation() {
-    await this.storage.remove(authCommon.IMPERSONATION_ACTIVE_KEY);
-    await this.storage.remove(authCommon.IMPERSONATION_USER_KEY);
-    await this.storage.remove(authCommon.IMPERSONATION_REAL_USER_AUTH_KEY);
+  clearImpersonation() {
+    return this.storage.remove(authCommon.IMPERSONATION_ACTIVE_KEY).then(() =>
```
mpobrien: maybe `Promise.all` would work nicely here

jsflax: What's the advantage of `Promise.all` here?

mpobrien: It accomplishes the same thing, but the resulting code is smaller/less noisy:
```
return Promise.all([
    this.storage.remove(authCommon.IMPERSONATION_ACTIVE_KEY),
    this.storage.remove(authCommon. IMPERSONATION_REAL_USER_AUTH_KEY),
    this.storage.remove(authCommon. IMPERSONATION_USER_KEY),
])
```

Also, we don't actually care about enforcing a strict ordering of how these keys get deleted, so chaining them together with `.then` isn't necessary (and maybe misleading to someone reading the code).

src/util.js
```diff
@@ -83,14 +83,20 @@ function serviceResponse(service, { serviceName = service.serviceName, action, a
...
  * @memberof util
  * @returns {Object} An object of the form {name: ..., version: ...}, or null
  */
 function getPlatform() {
-  return platform ? platform : null;
+  return platform ? platform : _platform;
+}
+
+function setPlatform(customPlatform) {
+  platform = customPlatform;
```
mpobrien: relying on mutable global state in this module is very smelly, especially when it's the core SDK code reaching into a util module...
is it absolutely necessary to do this and/or is there a different approach?

jsflax: It is necessary to maintain a separate variable for platform, since `detect-browser` won't work on non-browsers (have we tested this on node?). That being said, it doesn't explicitly need to be tied to global state.

I made it global because platform shouldn't be mutable once it's set. `platform` should, in theory, be a static value. Can't really change the platform you're on. That being said, since Global Variables Are Bad™ (http://wiki.c2.com/?GlobalVariablesAreBad), there are other alternatives.

I think a creative alternative would be to lock the state once it's set via a tokenized dispatcher, e.g,:

```javascript
let isPlatformSet = false;
function setPlatform(customPlatform) {
    if (isPlatformSet) {
        throw Error("platform has already been set");
    }

    platform = customPlatform;
}
```

I suppose another option could be a singleton that must be initialized.

jsflax: With that said, a parameterized Singleton doesn't really work in practice, and would have to fallback on the some mechanisms.

mpobrien: `detect-browser` works in node:
```
$ node -e 'console.log(require("detect-browser"))'
{ name: 'node', version: '8.6.0' }
```

I still don't understand why `platform` would ever need to be modified at all. If there's a need to derive some kind of `customPlatform` option that overrides what `getPlatform` normally returns, it should be dealt with in the caller. It looks like the use case is so that you can provide a `customPlatform` option to the client constructor. Since the platform is only actually ever used in `getDeviceInfo` which is in turn used by the auth providers, it seems like those auth methods should just take a `customPlatform` option that flows downward from the client class.

src/auth/storage.js
```diff
@@ -28,21 +28,24 @@ class Storage {
...
+  if (storageType === 'localStorage') {
     if ((typeof window !== 'undefined') && 'localStorage' in window && window.localStorage !== null) {
       return new Storage(window.localStorage);
     }
-  } else if (type === 'sessionStorage') {
+  } else if (storageType === 'sessionStorage') {
     if ((typeof window !== 'undefined') && 'sessionStorage' in window && window.sessionStorage !== null) {
       return new Storage(window.sessionStorage);
     }
+  } else if (storageType == 'customStorage') { //eslint-disable-line eqeqeq
```
mpobrien: why no === here?

jsflax: When I was working with `flow`, it disallowed triple equals. I'll change it back

src/util.js
jsflax: I'm not sure detect-browser will work with react-native, as they supply their own module for platform: https://facebook.github.io/react-native/docs/platform-specific-code.html .

jsflax:
@mpobrien removed global platform. PTAL

src/auth/index.js
```diff
@@ -21,7 +21,7 @@ export default class Auth {
     this.client = client;
     this.rootUrl = rootUrl;
     this.codec = options.codec;
-    this.platform = options.platform !== null ? options.platform : _platform;
+    this.platform = options.platform !== undefined ? options.platform : _platform;
```
mpobrien: could probably just do `this.platform == options.platform || _platform` anyway?

src/client.js
```diff
@@ -7,7 +7,7 @@ import ServiceRegistry from './services';
 import * as common from './common';
 import ExtJSON from 'mongodb-extjson';
 import queryString from 'query-string';
-import { deprecate, } from './util';
+import { deprecate } from './util';
```
mpobrien: this is sort of unrelated to the purpose of this ticket but - if we're planning to do a breaking (major) version for this change, maybe this is the right time to just remove the deprecated function bindings in this class. thoughts? cc:  @edaniels

src/auth/index.js
```diff
@@ -178,103 +213,121 @@ export default class Auth {
...
+      let realUserAuth = JSON.parse(userAuth);
+      return this.set(realUserAuth);
+    }).then(() => this.clearImpersonation());
   }
 
   clearImpersonation() {
-    this.storage.remove(authCommon.IMPERSONATION_ACTIVE_KEY);
-    this.storage.remove(authCommon.IMPERSONATION_USER_KEY);
-    this.storage.remove(authCommon.IMPERSONATION_REAL_USER_AUTH_KEY);
+    Promise.all([
```
mpobrien: nit: looks like this was overlooked in the original code but I think we should just `return` this promise to the caller so that it functions properly in a promise chain

src/auth/index.js
```diff
@@ -20,8 +21,33 @@ export default class Auth {
     this.client = client;
     this.rootUrl = rootUrl;
     this.codec = options.codec;
-    this.storage = createStorage(options.storageType);
-    this.providers = createProviders(this);
+    this.platform = options.platform !== undefined ? options.platform : _platform;
```
mpobrien: nit: could use `||` instead of ternary here, since i think we probably want the same behavior with other falsy values too

src/client.js
mpobrien: We'll need to check all our docs/samples and make sure that any places where those deprecated functions are used get updated to the new stuff.

PR #98, opened 2018-01-02
STITCH-1037 JS SDK: Namespace local storage of auth info by client app ID

jsflax:


src/auth/storage.js
```diff
@@ -24,30 +24,35 @@ class MemoryStorage {
 }
 
 class Storage {
-  constructor(store) {
+  /**
+   * @param {Storage} store implementer of Storage interface
+   * @param {String} namespace appID to be used for namespacing
```
edaniels: [nit] `appID` -> `clientAppId`

src/auth/storage.js
```diff
@@ -24,30 +24,35 @@ class MemoryStorage {
...
+  */
+  constructor(store, namespace) {
     this.store = store;
+    this.namespace = namespace;
   }
 
-  get(key) { return new Promise(resolve => resolve(this.store.getItem(key))); }
-  set(key, value) { return new Promise(resolve => resolve(this.store.setItem(key, value))); }
-  remove(key) { return new Promise(resolve => resolve(this.store.removeItem(key))); }
+  get(key) { return new Promise(resolve => resolve(this.store.getItem(`${this.namespace}.${key}`))); }
```
edaniels: [nit] It would be nice if there was a function to resolve the correct namespace

src/auth/storage.js
```diff
@@ -24,30 +24,35 @@ class MemoryStorage {
...
+  set(key, value) { return new Promise(resolve => resolve(this.store.setItem(`${this.namespace}.${key}`, value))); }
+  remove(key) { return new Promise(resolve => resolve(this.store.removeItem(`${this.namespace}.${key}`))); }
   clear() { return new Promise(resolve => resolve(this.store.clear())); }
```
edaniels: [q] Shouldn't clear just clear the specific namespace? Otherwise you cannot maintain multiple apps and log out of  one.

src/auth/storage.js
edaniels: [test] We should test that creating multiple apps works and that you can be logged in to two at the same time. As well as a test for logging out of one and being logged into the other.

src/auth/index.js
```diff
@@ -13,9 +13,17 @@ const EMBEDDED_USER_AUTH_DATA_PARTS = 4;
 
 export default class Auth {
   constructor(client, rootUrl, options) {
+    let namespace;
+    if (client === null || client.clientAppId === '') {
+      namespace = '<admin>';
```
edaniels: [nit] For safety purposes, what do you think of prefixing the client namespace with something like `client-${client.clientAppId}`? That way we can never mistake admin for client.

jsflax: Maybe `client.${client.clientAppId}`. Full stop might be easier to differentiate. I have no strong opinion either way.

edaniels: sure

src/auth/storage.js
jsflax: Good point, though this will be trickier than expected. `.clear` doesn't take any parameters– we'll have to keep a set of stored keys separately, and then iterate through each one to clear them out, never actually calling `.clear`.

edaniels: Actually looks like we don't call this clear method. Should we just remove it?

src/auth/index.js
```diff
@@ -13,9 +13,17 @@ const EMBEDDED_USER_AUTH_DATA_PARTS = 4;
 
 export default class Auth {
   constructor(client, rootUrl, options) {
+    let namespace;
+    if (client === null || client.clientAppId === '') {
```
mpobrien: nit: i think `if(!client || !client.clientAppId){ ... ` here might be a little more robust

src/auth/storage.js
jsflax: We call it in tests... it might not be a bad idea to have? In theory, we should be calling it on logout. I'll look into it.

jsflax:
```
clear() {
    return this.storage.remove(authCommon.USER_AUTH_KEY).then(() =>
      this.storage.remove(authCommon.REFRESH_TOKEN_KEY)
    ).then(() =>
      this.clearImpersonation
    );
  }
```

That is the `clear` method in the `Auth` class. Seems like it's trying to preserve some amount of saved data, which seems counter intuitive. Should we just be calling `this.storage.clear()` here?

edaniels: The device id should not be cleared. This looks correct if you look at the way Android does it

jsflax:
@edaniels PTAL

src/auth/storage.js
```diff
@@ -26,17 +26,75 @@ class MemoryStorage {
...
   constructor(store, namespace) {
     this.store = store;
     this.namespace = namespace;
+    this._keySetKey = `_${this.namespace}.keys`
+  }
+  
+  _generateKey(key) {
+    return `${this.namespace}.${key}`
+  }
+  _saveKeyToKeySet(generatedKey) {
```
edaniels: [minor] Instead of storing more into local storage, why don't we just keep track of what always needs to be cleared. It's only the two keys for user auth and refresh token. I think this increases the complexity a bit too much.

src/auth/index.js
```diff
@@ -14,10 +14,10 @@ const EMBEDDED_USER_AUTH_DATA_PARTS = 4;
 export default class Auth {
   constructor(client, rootUrl, options) {
     let namespace;
-    if (client === null || client.clientAppId === '') {
-      namespace = '<admin>';
+    if (!client || client.clientAppID === '') {
```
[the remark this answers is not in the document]
jsflax: There are tests that pass in null `client`s

jsflax:
I had thought this was going to be a "force logout" kind of situation according to the ticket:

> This will log out users and should be a maintenance SEMVER bump.

If we want to do a one time migration, I'm happy to add that as well.

src/auth/storage.js
jsflax: If that's the case, we can just go back to your original remark of just removing the `clear` function.

edaniels:
Yeah that's my fault. I realized we need to be much more fault tolerant for anonymous accounts and a jarring experience.

src/auth/storage.js
edaniels: 👍

src/auth/index.js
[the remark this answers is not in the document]
edaniels: but if `!client` is false, then it cannot be null?

jsflax:
Right on. Will add to the PR.

jsflax:
@edaniels Added the migration logic, and removed the clear function. PTAL

src/auth/storage.js
```diff
@@ -1,53 +1,124 @@
...
-  clear() {
-    this._data = {};
-    return this._data;
+  key(index) {
+    return this._orderedKeys[index];
+  }
+}
+
+const _VERSION = 1
+const _VERSION_KEY = '__storage_version__'
```
edaniels: [nit] prefix with the `_stitch` prefix

src/auth/storage.js
```diff
@@ -1,53 +1,124 @@
...
+    this._orderedKeys.push(key);
     this._data[key] = value;
+    this.length++;
     return this._data[key];
   }
 
   removeItem(key) {
+    this._orderedKeys.pop(key);
     delete this._data[key];
+    this.length--;
```
mpobrien: any reason we have to maintain `length` instead of just using `this._orderedKeys.length`?

edaniels:
I think it's safe to do the migration as an attempt. We unfortunately do not store the client app ID in the JWTs. You're right about the failure flow and this will only be happening to a select few clients that we've noticed are suffering from this issue.

src/auth/storage.js
jsflax: Yes– it is to fully implement the interface of the various other storage types: https://developer.mozilla.org/en-US/docs/Web/API/Storage/length . We need to give arbitrary accessors access to `.length` as a class member.

mpobrien: OK, makes sense. Alternately you could add a class method that does this with `get length(){ return this._orderedKeys.length; }`

PR #101, opened 2018-01-11
STITCH-1039 JS SDK: Improve semantics of multiple log ins

jsflax:


src/auth/index.js
```diff
@@ -67,7 +67,8 @@ export default class Auth {
   }
 
   refreshToken() {
-    return this.client.doSessionPost().then(json => this.set(json));
+    return Promise.all([this.client.doSessionPost(), this.getLoggedInProviderType()])
```
edaniels: [q] Why does refreshing a token set the provider type?

src/auth/index.js
```diff
@@ -123,7 +124,7 @@ export default class Auth {
       }
 
       // If we get here, the state is valid - set auth appropriately.
-      return this.set(redirectState.ua);
+      return this.getLoggedInProviderType().then(providerType => this.set(redirectState.ua, providerType));
```
edaniels: [major] How does the ordering work here? `getLoggedInProviderType` wouldn't be able to return anything useful when it's the first time we're here and as a result, `set` will fail due to the restriction it has in place about empty types.

src/auth/index.js
```diff
@@ -160,14 +161,18 @@ export default class Auth {
 
     document.cookie = `${authCommon.USER_AUTH_COOKIE_NAME}=; Path=/; Expires=Thu, 01 Jan 1970 00:00:00 GMT;`;
     const userAuth = this.unmarshallUserAuth(uaCookie);
-    return this.set(userAuth).then(() =>
+    return this.getLoggedInProviderType().then(providerType => this.set(userAuth, providerType)).then(() =>
```
edaniels: [major] How does the ordering work here? `getLoggedInProviderType` wouldn't be able to return anything useful when it's the first time we're here and as a result, `set` will fail due to the restriction it has in place about empty types.

src/auth/index.js
```diff
@@ -208,19 +213,21 @@ export default class Auth {
     return this.storage.get(authCommon.REFRESH_TOKEN_KEY);
   }
 
-  set(json) {
-    if (!json) {
+  set(json, authType) {
+    if (!json || !authType) {
```
edaniels: It looks like there are cases where `authType` could definitely be empty which would cause us to not complete a log in. Is it necessary that we care what `authType` is?

src/auth/providers.js
```diff
@@ -7,6 +7,8 @@ import { uriEncodeObject } from '../util';
  * @namespace
  */
 function anonProvider(auth) {
+  const providerType = 'anon';
```
edaniels: [nit] We should probably make this exactly the same name/type that we use in the route for every provider. i.e. `anon` -> `anon-user`.

src/client.js
```diff
@@ -136,14 +136,29 @@ export default class StitchClient {
...
+    ).then(([accessToken, loggedInProviderType]) => {
       if (accessToken) {
-        return this.auth.authedId();
+        // TODO: `anon` should be enumerated once providers are maintain
+        // an interface
+        if (providerType === 'anon' && loggedInProviderType === 'anon') {
+          return false; // is authenticated, skip log in
+        }
+
+        return this.logout(); // will not be authenticated, continue log in
```
edaniels: [minor] If logout fails, then we will never authenticate. From ticket:
`the SDK should attempt to log out (failure should be swallowed), and logged in with the new provider`. We should have a test case for this.

src/client.js
```diff
@@ -136,14 +136,29 @@ export default class StitchClient {
...
-      return this.auth.provider(providerType).authenticate(options)
-        .then(() => this.auth.authedId());
-    });
+      return true; // is not authenticated, continue log in
+    }).then((isNotAuthed) => {
+      if (isNotAuthed) {
+        return this.auth.provider(providerType).authenticate(options);
+      }
+      return;
+    }).then(() => this.auth.authedId());
```
edaniels: [q] Can just pass `this.auth.authedId` instead of having an arrow function, right?

test/auth.test.js
```diff
@@ -26,18 +29,24 @@ function mockApiResponse(options = {}) {
...
+  });
 
   it('should return a promise for anonymous login with existing auth data', async() => {
     window.fetch.resolves(mockApiResponse());
     expect.assertions(1);
 
     let client = new StitchClient();
     await client.auth.storage.set(common.USER_AUTH_KEY, mockAuthData());
-
+    await client.auth.storage.set(common.USER_LOGGED_IN_PT_KEY, 'anon');
```
edaniels: [nit] This should probably use a constant

test/auth.test.js
```diff
@@ -58,15 +67,48 @@ describe('Auth', () => {
...
+  beforeEach(async() => {
+    const { apiKey, groupId, serverUrl } = extractTestFixtureDataPoints(test);
+    th = await buildClientTestHarness(apiKey, groupId, serverUrl);
+    await th.configureAnon();
+    client = th.stitchClient;
+  });
+
+  afterEach(async() => await th.cleanup());
+
+  it('should track currently logged in provider type', async() => {
```
edaniels: [minor] Per the ticket, we need a `isAuthenticated` like method to maintain parity with other SDKs. We should have a test that checks a flow of:
1. Verify `isAuthenticated` resolves to false
2. Log in
3. Verify `isAuthenticated` resolves to true

edaniels: [minor] We should assert a logout did indeed happen besides just re logging in.

src/client.js
```diff
@@ -136,14 +136,29 @@ export default class StitchClient {
...
-    return this.auth.getAccessToken().then(accessToken => {
+    return Promise.all(
+      [
+        this.auth.getAccessToken(),
+        this.auth.getLoggedInProviderType()
+      ]
+    ).then(([accessToken, loggedInProviderType]) => {
       if (accessToken) {
-        return this.auth.authedId();
+        // TODO: `anon` should be enumerated once providers are maintain
```
edaniels: [nit] Instead of enumeration, we could just have a constant for now

src/auth/providers.js
```diff
@@ -292,7 +298,7 @@ function googleProvider(auth) {
         })
           .then(common.checkStatus)
           .then(response => response.json())
-          .then(json => auth.set(json));
+          .then(json => auth.set(json, providerType));
       }
 
       const redirectUrl = (data && data.redirectUrl) ? data.redirectUrl : undefined;
```
edaniels: [minor] Redirects are not being handled for provider types being prospectively set in advance. You may be able to set a temporary field related to the request like STATE_KEY

edaniels: [major] Speaking of `STATE_KEY`, I'm pretty convinced `window.location.replace(getOAuthLoginURL(auth, 'google', redirectUrl));` and the facebook one are broken with the async changes to storage. getOAuthLoginURL needs to be thenned before using window location replace.

src/auth/index.js
edaniels: This one isn't like redirect and in this special case, we can assume the providerType is `mongodb-cloud`.

src/client.js
```diff
@@ -136,14 +136,29 @@ export default class StitchClient {
    */
   authenticate(providerType, options = {}) {
     // reuse existing auth if present
-    return this.auth.getAccessToken().then(accessToken => {
+    return Promise.all(
+      [
+        this.auth.getAccessToken(),
```
edaniels: [nit] It seems like it would be better to use just `getLoggedInProviderType` here right? You're only acting upon info from `providerType`

edaniels:
Evergreen tests are failing. Might be legit

src/auth/index.js
```diff
@@ -264,6 +271,10 @@ export default class Auth {
     });
   }
 
+  getLoggedInProviderType() {
+    return this.storage.get(authCommon.USER_LOGGED_IN_PT_KEY);
```
edaniels: [minor] Given that this directly proxies a storage async method, we should probably be better about the resolved output regardless of failure conditions. Under normal circumstances, this should return the type as a string, otherwise, an empty string could suffice. Another option is to throw an exception if `isAuthenticated` resolves to false.

src/client.js
edaniels: Now that I think about it, it seems like we actually want to check `isAuthenticated` and then `getLoggedInProviderType`. The thing to keep in mind is that we only want to call `getLoggedInProviderType` if it's going to not reject or anything. Otherwise we drop out of the next then and never authenticate.

src/auth/index.js
jsflax: What is a case where `authType` could be empty? We don't care what `authType` is, but we care if it's `null`. It should be impossible to log in to stitch without a provider type.

src/auth/index.js
jsflax: An exception could make sense, otherwise I think an `undefined` or `null` is an accurate representation of there not being a provider type.

src/client.js
jsflax: `isAuthenticated` doesn't exist exactly– it would be a mix of checking these fields anyway.

src/auth/index.js
edaniels: The handling of cookies and redirects

src/client.js
jsflax: We could, but they'd exist in an exported global state. Otherwise we'd have to heavily refactor `providers`

src/client.js
```diff
@@ -160,7 +175,8 @@ export default class StitchClient {
         useRefreshToken: true,
         rootURL: this.rootURLsByAPIVersion[v2][API_TYPE_CLIENT]
       }
-    ).then(() => this.auth.clear()); // eslint-disable-line space-before-function-paren
+    ).then(() => this.auth.clear())
+      .then(() => true);
```
mpobrien: q: why is this needed?

src/client.js
edaniels: I think it should exist. Users need to have a decent way of checking if they are authed before doing anything. authedId seems roundabout

test/auth.test.js
```diff
@@ -26,18 +29,24 @@ function mockApiResponse(options = {}) {
...
 describe('Auth', () => {
-  beforeEach(() => { test.fetch = sinon.stub(window, 'fetch'); });
-  afterEach(() => test.fetch.restore());
+  let test = {};
+
+  beforeEach(() => {
+    test.fetch = sinon.stub(window, 'fetch');
+  });
+
+  afterEach(async() => {
```
mpobrien: nit: this doesn't need to be `async` if we're not using `await` in the body

src/client.js
edaniels: Also, consider someone with existing local storage or a new SDK.  Won't `authenticate` ultimately reject because they have nothing set.

src/client.js
jsflax: Probably. Will have to try.

test/auth.test.js
jsflax: I assert this in a later test.

jsflax: `isAuthenticated` is going to be a wrapper over checking `authedId`. It seems weird to have multiple sources of truth for this, but I can add the method.

src/client.js
edaniels: Why would it be global if you're just importing the definitions?

src/auth/index.js
edaniels: Yeah so if we make it an exception, we will have to be very careful using it with Promise.all by not using it all and instead chaining it with `isAuthenticated` and then `getLoggedInProviderType`. `undefined` or `null` is okay too.

src/auth/providers.js
edaniels: I take this back. We should really haver a constant here that is used in createProviders as well.

test/auth.test.js
edaniels: I don't see that the other test actually asserts that we sent a logout to the server instead of re logging in.

edaniels: I'd argue that `authedId` should throw if you're not authed.

src/auth/providers.js
```diff
@@ -267,6 +272,7 @@ function getOAuthLoginURL(auth, providerName, redirectUrl) {
 /** @namespace */
 function googleProvider(auth) {
   const loginRoute = auth.isAppClient() ? 'providers/oauth2-google/login' : 'providers/oauth2-google/login';
+  const providerType = 'google';
```
mpobrien: we should declare these special strings in constants up top. we use em all over (including in tests)

test/auth.test.js
```diff
@@ -58,15 +67,48 @@ describe('Auth', () => {
...
+    await client.login(email, password);
+    expect(await client.auth.storage.get(common.USER_LOGGED_IN_PT_KEY)).toEqual('userpass');
+  });
+
+  it('should return existing login for relogging "anon"', async() => {
+    const userId = await client.login();
+    expect(userId).toEqual(await client.login());
+  });
+
+  it('should login with a new provider if switching providers', async() => {
```
mpobrien: it looks like we sorta already do this in one of the previous tests at L94-95. Maybe we should just nest this `it` block into there?

src/auth/index.js
jsflax: These functions were locked together and would've required a larger refactor.

edaniels: Gotcha 👍

src/client.js
jsflax: I'll add a test– the code above should work to spec.

src/client.js
jsflax: As a successful return value. `clear()` doesn't produce anything useful– it would be nice to have a primitive value to check on the success of the logout.

test/auth.test.js
jsflax: Didn't assert, but it's checked implicitly in `should re-login with the same provider if not anon`. I'll add the assertion.

test/auth.test.js
jsflax: Sure, I don't have a strong opinion here. This test was just more explicit.

test/auth.test.js
jsflax: Good catch, leftover artifact from an earlier version of this.

src/client.js
jsflax: Right, I guess just because it isn't bound to any class or function doesn't mean it's global if it's exported properly and is stateless. I'm fine with making these constants, but it will be much nicer when each provider is implementing an interface that owns this const.

src/client.js
mpobrien: but the logout is implicitly successful if the promise that it returns doesn't reject, which the caller can check for with `.then`/`.catch`. It doesn't look like we use the `true` value anyway, so i don't see the point

src/client.js
jsflax: Apparently not, unless I'm doing it wrong.

src/client.js
jsflax: We use it in the new authenticate method.

src/client.js
edaniels: nah makes sense, the this here is the client

src/client.js
```diff
@@ -136,14 +137,27 @@ export default class StitchClient {
...
+
+        return this.logout(); // will not be authenticated, continue log in
       }
 
-      return this.auth.provider(providerType).authenticate(options)
-        .then(() => this.auth.authedId());
-    });
+      return true; // is not authenticated, continue log in
+    }).then((isNotAuthed) => {
+      if (isNotAuthed) {
```
edaniels: [nit] Feels like the name of `isNotAuthed` is more apt as `shouldAuth`.

src/client.js
```diff
@@ -160,7 +174,8 @@ export default class StitchClient {
         useRefreshToken: true,
         rootURL: this.rootURLsByAPIVersion[v2][API_TYPE_CLIENT]
       }
-    ).then(() => this.auth.clear()); // eslint-disable-line space-before-function-paren
+    ).then(() => this.auth.clear())
```
edaniels: [minor] I think we need a `catch` or a second arg to `then` such that in both cases we succeed and always return with `this.auth.clear`. We can log the failure.

src/client.js
```diff
@@ -136,14 +137,27 @@ export default class StitchClient {
...
+        this.auth.getLoggedInProviderType()
+      ]
+    ).then(([accessToken, loggedInProviderType]) => {
       if (accessToken) {
-        return this.auth.authedId();
+        if (providerType === PROVIDER_TYPE_ANON && loggedInProviderType === PROVIDER_TYPE_ANON) {
+          return false; // is authenticated, skip log in
+        }
+
+        return this.logout(); // will not be authenticated, continue log in
```
edaniels: [minor] I think logout we really don't care if it fails or not, just that we try and wait to do it. Given my comment on the logout method holds, we can just do `this.logout().then(() => true)`

src/client.js
```diff
@@ -183,6 +198,14 @@ export default class StitchClient {
     )
       .then(response => response.json());
   }
+
+  /**
+  * @return {Promise} whether or not the current client is authenticated
+  */
+  isAuthenticated() {
+    return this.auth.authedId().then((id) => id != null && id != undefined);
```
edaniels: [minor] This promise can reject. If it's suppose to return a bool, we should probably satisfy the `reject` case on `then`.
[test] Test for when it rejects

test/auth.test.js
```diff
@@ -1,6 +1,11 @@
 import sinon from 'sinon';
 import StitchClient from '../src/client';
 import * as common from '../src/auth/common';
+import { PROVIDER_TYPE_ANON, PROVIDER_TYPE_USERPASS } from '../src/auth/providers';
+
+const StitchMongoFixture = require('./fixtures/stitch_mongo_fixture');
```
edaniels: [nit] `import StitchMongoFixture from './fixtures/stitch_mongo_fixture'`
[style] collapse lines

test/auth.test.js
```diff
@@ -58,15 +67,57 @@ describe('Auth', () => {
...
+describe('Auth login semantics', () => {
+  const email = 'test_user@domain.com';
+  const password = 'password';
+  let test = new StitchMongoFixture();
+  let th;
+  let client;
 
-    let client = new StitchClient();
-    await client.auth.storage.set(common.USER_AUTH_KEY, mockAuthData());
+  beforeAll(async() => await test.setup());
```
edaniels: [q] If `test.setup` is a promise, can't you just do `beforeAll(test.setup())`?

test/auth.test.js
```diff
@@ -58,15 +67,57 @@ describe('Auth', () => {
...
+  afterAll(async() => await test.teardown());
```
edaniels: Same question to all uses of before* and after* where it's a single line return

src/auth/index.js
```diff
@@ -208,19 +213,28 @@ export default class Auth {
...
   }
 
-  set(json) {
+  setAuthType(authType) {
+    return new Promise((resolve) =>
+      resolve(this.storage.set(authCommon.USER_LOGGED_IN_PT_KEY, authType))
+    );
+  }
+
+  set(json, authType) {
```
mburdette: you could use a default value `authType = ''` and then not do the `!authType` check in the promise that @edaniels commented on

src/auth/index.js
```diff
@@ -264,6 +278,10 @@ export default class Auth {
     });
   }
 
+  getLoggedInProviderType() {
+    return this.storage.get(authCommon.USER_LOGGED_IN_PT_KEY).then(type => !type ? '' : type);
```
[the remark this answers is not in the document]
mburdette: [nit] even easier would be `type || ''`

src/auth/index.js
```diff
@@ -208,19 +213,28 @@ export default class Auth {
...
+
+  set(json, authType) {
     if (!json) {
       return;
     }
 
     let newUserAuth = {};
     return new Promise((resolve) => {
+      if (!authType) { resolve(); }
+      else { resolve(this.storage.set(authCommon.USER_LOGGED_IN_PT_KEY, authType)); }
```
mpobrien: I think you could just use `return` statements here and not need to involve `resolve` altogether.

src/client.js
```diff
@@ -136,14 +137,27 @@ export default class StitchClient {
...
 
-      return this.auth.provider(providerType).authenticate(options)
-        .then(() => this.auth.authedId());
-    });
+      return true; // is not authenticated, continue log in
+    }).then((isNotAuthed) => {
+      if (isNotAuthed) {
+        return this.auth.provider(providerType).authenticate(options);
+      }
+      return;
```
mpobrien: superfluous `return` here

src/auth/index.js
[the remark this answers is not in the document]
edaniels: ^ If you're a masochist

src/client.js
```diff
@@ -136,14 +137,27 @@ export default class StitchClient {
    */
   authenticate(providerType, options = {}) {
     // reuse existing auth if present
-    return this.auth.getAccessToken().then(accessToken => {
+    return Promise.all(
+      [
+        this.auth.getAccessToken(),
```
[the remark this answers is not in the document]
jsflax: I didn't realize it rejected if nil. I'll move it around.

src/client.js
jsflax: good point. will do

src/client.js
jsflax: Fair point.

src/client.js
jsflax: Good call

test/auth.test.js
jsflax: Yep, will do.

src/auth/index.js
```diff
@@ -208,19 +213,28 @@ export default class Auth {
...
+  }
+
+  set(json, authType) {
     if (!json) {
       return;
     }
 
     let newUserAuth = {};
     return new Promise((resolve) => {
+      if (!authType) { resolve(); }
```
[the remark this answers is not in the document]
jsflax: Will address with @mburdette 's comment

test/testutil.js
```diff
@@ -51,6 +51,12 @@ class TestHarness {
     });
   }
 
+  async configureAnon() {
```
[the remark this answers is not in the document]
jsflax: will remove

test/auth.test.js
jsflax: fixing

src/auth/index.js
jsflax: done

src/auth/index.js
jsflax: done

src/client.js
jsflax: done

test/auth.test.js
jsflax: Apparently, that isn't allowed.

test/auth.test.js
jsflax: Apparently, that isn't allowed.

edaniels: why not?

jsflax:
```
encountered a declaration exception
TypeError: fn.call is not a function
      
      at Object.asyncFn (node_modules/jest-jasmine2/build/jasmine-async.js:38:30)
          at Promise (<anonymous>)
      at Promise.resolve.then.el (node_modules/p-map/index.js:46:16)
          at <anonymous>
      at process._tickCallback (internal/process/next_tick.js:188:7)
```

It's expecting to be wrapped in a block by the looks of it

jsflax:
@edaniels @mpobrien @mburdette PTAL

test/auth.test.js
edaniels: what code are you passing it?

src/auth/index.js
```diff
@@ -214,27 +214,25 @@ export default class Auth {
...
 
     let newUserAuth = {};
-    return new Promise((resolve) => {
-      if (!authType) { resolve(); }
-      else { resolve(this.storage.set(authCommon.USER_LOGGED_IN_PT_KEY, authType)); }
-    }).then(() => {
+
+    return (authType ? 
+      this.storage.set(authCommon.USER_LOGGED_IN_PT_KEY, authType) : 
+      new Promise()
```
mpobrien: I think this needs to be `Promise.resolve()`, rather than `new Promise()`.

test/auth.test.js
jsflax: `beforeAll(test.setup())` / the equivalent for each block

src/client.js
mpobrien: i think it's still there!

src/auth/index.js
jsflax: `new Promise()` has been tested and works. But if you'd prefer `Promise.resolve()`, I can alter it.

test/auth.test.js
edaniels: darn. oh well

src/auth/index.js
mpobrien: Are you sure? I think you're required to pass a function to the `Promise` constructor otherwise there's no way for it to hook into the following `.then`/`.catch`.
e.g. if i do that I get this:
```
> (new Promise()).then(()=>{console.log("hi")})
TypeError: Promise resolver undefined is not a function
    at Promise (<anonymous>)
```

if i use `Promise.resolve` though i get:
```> (Promise.resolve()).then(()=>{console.log("hi")})
Promise {
  <pending>,
  domain:
   Domain {
     domain: null,
     _events: { error: [Function: debugDomainError] },
     _eventsCount: 1,
     _maxListeners: undefined,
     members: [] } }
> hi
```

jsflax: Fixed

src/auth/index.js
```diff
@@ -67,7 +67,8 @@ export default class Auth {
   }
 
   refreshToken() {
-    return this.client.doSessionPost().then(json => this.set(json));
+    return Promise.all([this.client.doSessionPost(), this.getLoggedInProviderType()])
+      .then(([json, providerType]) => this.set(json, providerType));
```
edaniels: Remove `Promise.all`

PR #107, opened 2018-01-22
STITCH-1122: Create helper method for findOne() in JS SDK

BlakeIsBlake:


BlakeIsBlake:
Did this the right way RFAL

mpobrien:
👍 LGTM

PR #108, opened 2018-01-23
STITCH-1121 JS SDK: Add static Create client function to avoid async …

jsflax:
…calls

### DIFF
--- a/src/admin.js
+++ b/src/admin.js
@@ -1,14 +1,26 @@
 /* global window, fetch */
 /* eslint no-labels: ['error', { 'allowLoop': true }] */
 import 'fetch-everywhere';
-import StitchClient from './client';
+import { _StitchClient } from './client';
 import ADMIN_CLIENT_TYPE from './common';
 import { ADMIN_CLIENT_CODEC } from './auth/common';
 
 const v2 = 2;
 const v3 = 3;
 
-export default class Admin extends StitchClient {
+export class Admin {
+  constructor() {
+    throw new StitchError('Constructor undefined. Use `Admin.init` instead.');
+  }
+
+  static init(baseUrl) {
+    return new Promise((resolve, reject) => {
+      resolve(new _Admin(baseUrl));
+    });
+  }
+}
+
+export class _Admin extends _StitchClient {
   constructor(baseUrl) {
     super('', {baseUrl, authCodec: ADMIN_CLIENT_CODEC});
   }
--- a/src/admin.js
+++ b/src/admin.js
@@ -8,9 +8,9 @@ import { ADMIN_CLIENT_CODEC } from './auth/common';
 const v2 = 2;
 const v3 = 3;
 
-export class StitchAdminFactory {
+export class StitchAdminClientFactory {
   constructor() {
-    throw new StitchError('StitchAdminClient can only be made from the StitchAdminFactory.create function');
+    throw new StitchError('StitchAdminClient can only be made from the StitchAdminClientFactory.create function');
   }
 
   static create(baseUrl) {
--- a/src/admin.js
+++ b/src/admin.js
@@ -1,7 +1,7 @@
 /* global window, fetch */
 /* eslint no-labels: ['error', { 'allowLoop': true }] */
 import 'fetch-everywhere';
-import { StitchClient } from './client';
+import { newStitchClient, StitchClient } from './client';
 import ADMIN_CLIENT_TYPE from './common';
 import { ADMIN_CLIENT_CODEC } from './auth/common';
 
@@ -14,15 +14,14 @@ export class StitchAdminClientFactory {
   }
 
   static create(baseUrl) {
-    return new Promise((resolve, reject) => {
-      resolve(new StitchAdminClient(baseUrl)._willInitialize);
-    });
+    return newStitchClient(StitchAdminClient.prototype, '', {baseUrl, authCodec: ADMIN_CLIENT_CODEC});
   }
 }
 
 export class StitchAdminClient extends StitchClient {
-  constructor(baseUrl) {
-    super('', {baseUrl, authCodec: ADMIN_CLIENT_CODEC});
+  constructor() {
+    super();
+    throw new StitchError('StitchAdminClient can only be made from the StitchAdminClientFactory.create function');
   }
 
   get type() {
--- a/src/admin.js
+++ b/src/admin.js
@@ -1,26 +1,26 @@
 /* global window, fetch */
 /* eslint no-labels: ['error', { 'allowLoop': true }] */
 import 'fetch-everywhere';
-import { _StitchClient } from './client';
+import { StitchClient } from './client';
 import ADMIN_CLIENT_TYPE from './common';
 import { ADMIN_CLIENT_CODEC } from './auth/common';
 
 const v2 = 2;
 const v3 = 3;
 
-export class Admin {
+export class StitchAdminFactory {
   constructor() {
-    throw new StitchError('Constructor undefined. Use `Admin.init` instead.');
+    throw new StitchError('StitchAdminClient can only be made from the StitchAdminFactory.create function');
   }
 
-  static init(baseUrl) {
+  static create(baseUrl) {
     return new Promise((resolve, reject) => {
-      resolve(new _Admin(baseUrl));
+      resolve(new StitchAdminClient(baseUrl)._willInitialize);
     });
   }
 }
 
-export class _Admin extends _StitchClient {
+export class StitchAdminClient extends StitchClient {
   constructor(baseUrl) {
     super('', {baseUrl, authCodec: ADMIN_CLIENT_CODEC});
   }
--- a/src/auth/index.js
+++ b/src/auth/index.js
@@ -11,29 +11,45 @@ const jwtDecode = require('jwt-decode');
 
 const EMBEDDED_USER_AUTH_DATA_PARTS = 4;
 
-export default class Auth {
+export class AuthFactory {
+  constructor() {
+    throw new StitchError('StitchClient can only be made from the StitchClientFactory.create function');
+  }
+
+  static create(client, rootUrl, options) {
+    return new Promise((resolve, reject) => {
+      resolve(new Auth(client, rootUrl, options)._willInitialize);
+    });
+  }
+}
+
+export class Auth {
   constructor(client, rootUrl, options) {
-    let namespace;
-    if (!client || client.clientAppID === '') {
-      namespace = 'admin';
-    } else {
-      namespace = `client.${client.clientAppID}`;
-    }
+    this._willInitialize = new Promise((resolve, reject) => {
+      let namespace;
+      if (!client || client.clientAppID === '') {
+        namespace = 'admin';
+      } else {
+        namespace = `client.${client.clientAppID}`;
+      }
 
-    options = Object.assign({
-      codec: authCommon.APP_CLIENT_CODEC,
-      namespace: namespace,
-      storageType: 'localStorage'
-    }, options);
+      options = Object.assign({
+        codec: authCommon.APP_CLIENT_CODEC,
+        namespace: namespace,
+        storageType: 'localStorage'
+      }, options);
 
-    this.client = client;
-    this.rootUrl = rootUrl;
-    this.codec = options.codec;
-    this.platform = options.platform || _platform;
-    this.storage = createStorage(options);
-    this.providers = createProviders(this, options);
-    this._willInitialize = this._get().then(auth => {
-      this.authedId = auth.userId;
+      this.client = client;
+      this.rootUrl = rootUrl;
+      this.codec = options.codec;
+      this.platform = options.platform || _platform;
+      this.storage = createStorage(options);
+      this.providers = createProviders(this, options);
+      this._get().then(auth => {
+        this.authedId = auth.userId;
+        this._isInitialized = true;
+        resolve(this);
+      });
     });
   }
 
--- a/src/auth/index.js
+++ b/src/auth/index.js
@@ -42,12 +42,12 @@ export function newAuth(client, rootUrl, options) {
   auth.platform = options.platform || _platform;
   auth.storage = createStorage(options);
   auth.providers = createProviders(auth, options);
-  
+
   return Promise.all([
-      auth._get(),
-      auth.storage.get(authCommon.REFRESH_TOKEN_KEY),
-      auth.storage.get(authCommon.USER_LOGGED_IN_PT_KEY),
-      auth.storage.get(authCommon.DEVICE_ID_KEY)
+    auth._get(),
+    auth.storage.get(authCommon.REFRESH_TOKEN_KEY),
+    auth.storage.get(authCommon.USER_LOGGED_IN_PT_KEY),
+    auth.storage.get(authCommon.DEVICE_ID_KEY)
   ]).then(([authObj, rt, loggedInProviderType, deviceId]) => {
     auth.auth = authObj;
     auth.authedId = authObj.userId;
--- a/src/auth/index.js
+++ b/src/auth/index.js
@@ -13,55 +13,57 @@ const EMBEDDED_USER_AUTH_DATA_PARTS = 4;
 
 export class AuthFactory {
   constructor() {
-    throw new StitchError('StitchClient can only be made from the StitchClientFactory.create function');
+    throw new StitchError('Auth can only be made from the AuthFactory.create function');
   }
 
   static create(client, rootUrl, options) {
-    return new Promise((resolve, reject) => {
-      resolve(new Auth(client, rootUrl, options)._willInitialize);
-    });
+    return newAuth(client, rootUrl, options);
   }
+}
+
+export function newAuth(client, rootUrl, options) {
+  return new Promise((resolve, reject) => {
+    let auth = Object.create(Auth.prototype);
+    let namespace;
+    if (!client || client.clientAppID === '') {
+      namespace = 'admin';
+    } else {
+      namespace = `client.${client.clientAppID}`;
+    }
+
+    options = Object.assign({
+      codec: authCommon.APP_CLIENT_CODEC,
+      namespace: namespace,
+      storageType: 'localStorage'
+    }, options);
+
+    auth.client = client;
+    auth.rootUrl = rootUrl;
+    auth.codec = options.codec;
+    auth.platform = options.platform || _platform;
+    auth.storage = createStorage(options);
+    auth.providers = createProviders(auth, options);
+
+    Promise.all([
+      auth._get(),
+      auth.storage.get(authCommon.REFRESH_TOKEN_KEY),
+      auth.storage.get(authCommon.USER_LOGGED_IN_PT_KEY),
+      auth.storage.get(authCommon.DEVICE_ID_KEY)
+    ]).then(([authObj, rt, loggedInProviderType, deviceId]) => {
+      auth.auth = authObj;
+      auth.authedId = authObj.userId;
+      auth.rt = rt;
+      auth.loggedInProviderType = loggedInProviderType;
+      auth.deviceId = deviceId;
+
+      resolve(auth);
+    });
+  });
 }
 
 export class Auth {
   constructor(client, rootUrl, options) {
-    this._willInitialize = new Promise((resolve, reject) => {
-      let namespace;
-      if (!client || client.clientAppID === '') {
-        namespace = 'admin';
-      } else {
-        namespace = `client.${client.clientAppID}`;
-      }
-
-      options = Object.assign({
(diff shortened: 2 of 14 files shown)


src/admin.js
```diff
@@ -1,14 +1,26 @@
...
 
-export default class Admin extends StitchClient {
+export class Admin {
+  constructor() {
+    throw new StitchError('Constructor undefined. Use `Admin.init` instead.');
+  }
+
+  static init(baseUrl) {
+    return new Promise((resolve, reject) => {
+      resolve(new _Admin(baseUrl));
```
edaniels: Based on my other comment, I think this can call some "private"/"unexported" method to construct the client. I imagine we'd call `StitchClient.init()`

src/client.js
```diff
@@ -23,12 +23,29 @@ const API_TYPE_CLIENT = 'client';
 const API_TYPE_APP = 'app';
 
 /**
- * Create a new StitchClient instance.
+  * Factory class to create a new StitchClient asynchronously.
+  */
+export class StitchClient {
```
edaniels: [minor] Same comment about having just one combined class that we export as we discussed earlier.

src/client.js
```diff
@@ -23,12 +23,29 @@ const API_TYPE_CLIENT = 'client';
...
+ * Create a new _StitchClient instance.
+ * This is the internal implementation for StitchClient and should not
+ * be exposed.
  *
  * @class
  * @return {StitchClient} a StitchClient instance.
  */
-export default class StitchClient {
+export class _StitchClient {
   constructor(clientAppID, options = {}) {
```
edaniels: This is probably what should become the private function and not the actual constructor. This can return a promise itself and not have to hold any vars regarding initialization like `_isInitialized` and `_isInitialized`.

src/client.js
```diff
@@ -199,17 +221,18 @@ export default class StitchClient {
   }
 
   /**
-  * @return {Promise} whether or not the current client is authenticated
+  * @return {Boolean} whether or not the current client is authenticated
   */
   isAuthenticated() {
-    return this.auth.authedId().then((id) => id !== null && id !== undefined, () => false);
+    const authedId = this.authedId();
```
edaniels: [nit] `!!this.authedId` would probably suffice

src/client.js
```diff
@@ -199,17 +221,18 @@ export default class StitchClient {
...
+    return authedId !== null && authedId !== undefined;
   }
 
   /**
-   *  @return {Promise} Returns a promise resolving to a string of the currently authed user's ID.
+   *  @return {String} Returns a promise resolving to a string of the currently authed user's ID.
```
edaniels: [typo] returns the currently...

src/client.js
```diff
@@ -464,35 +487,33 @@ export default class StitchClient {
 
     let { url, fetchArgs } = this._fetchArgs(resource, method, options);
     if (!options.noAuth) {
-      return this.authedId().then(authedId => {
-        if (!this.authedId) {
-          return Promise.reject(new StitchError('Must auth first', ErrUnauthorized));
-        }
+      if (!this.authedId()) {
```
edaniels: [nit] The more appropriate check is `!this.isAuthenticated()`

src/client.js
```diff
@@ -464,35 +487,33 @@ export default class StitchClient {
...
-              return this.auth.refreshToken().then(() => {
-                options.refreshOnFailure = false;
-                return this._do(resource, method, options);
-              });
-            }
-
-            return tokenPromise.then(token => {
-              fetchArgs.headers.Authorization = `Bearer ${token}`;
-              return this._fetch(url, fetchArgs, resource, method, options);
+      let tokenPromise =
```
edaniels: [q] Why is this a promise still? Getting refresh token and access token should be a part of initialization

src/client.js
```diff
@@ -464,35 +487,33 @@ export default class StitchClient {
...
+        options.useRefreshToken ? this.auth.getRefreshToken() : this.auth.getAccessToken();
+
+      // If access token is expired, proactively get a new one
+      if (!options.useRefreshToken) {
+        return this.auth.isAccessTokenExpired().then(isAccessTokenExpired => {
```
edaniels: [q] `isAccessTokenExpired` cool be synchronous now too I think, right?

src/client.js
```diff
@@ -23,12 +23,29 @@ const API_TYPE_CLIENT = 'client';
...
+  * Factory class to create a new StitchClient asynchronously.
+  */
+export class StitchClient {
+  constructor() {
+    throw new StitchError('Constructor undefined. Use `StitchClient.init` instead.');
+  }
+
+  static init(clientAppID, options = {}) {
+    return new Promise((resolve, reject) => {
+      resolve(new _StitchClient(clientAppID, options));
```
edaniels: I don't see how this will actually wait for auth to initialize since this is still calling a constructor

src/auth/index.js
```diff
@@ -32,6 +32,9 @@ export default class Auth {
     this.platform = options.platform || _platform;
     this.storage = createStorage(options);
     this.providers = createProviders(this, options);
+    this._willInitialize = this._get().then(auth => {
```
edaniels: It seems like you may want a similar pattern for a static method on Auth to initialize an instance of it with no constructor such that we can get rid of these initialization variables

src/auth/index.js
```diff
@@ -264,6 +268,7 @@ export default class Auth {
       return this._get();
     }).then((auth) => {
       newUserAuth = Object.assign(auth, newUserAuth);
```
edaniels: [minor] Seems like there are many other methods here that could be made synchronous since we can load data for refresh, access token, and so on up front. The only time we would need to be async would be for initialization, making a network request, and setting new data. We don't strictly need to do this for everything, but we could. Piecemeal is fine as well. The Android SDK does the same right now but could load more up front.

test/auth.test.js
```diff
@@ -1,16 +1,16 @@
...
 import StitchMongoFixture from './fixtures/stitch_mongo_fixture';
 import { buildClientTestHarness, extractTestFixtureDataPoints } from './testutil';
 
 function mockAuthData() {
   const data = {
-    accessToken: 'fake-access-token',
-    refreshToken: 'fake-refresh-token',
-    userId: 'fake-user-id',
-    deviceId: 'fake-device-id'
+    access_token: 'fake-access-token',
```
edaniels: [q] Why did this need to change?

test/stitch_client.test.js
```diff
@@ -1,10 +1,52 @@
...
+    expect(stitchClient).toBeDefined();
+	  const authedId = await stitchClient.authenticate('userpass', th.userCredentials);
+	  expect(authedId).toBeDefined();
+  });
+
+  it('should allow closure access to a StitchClient', async() => {
+	  expect.assertions(2);
+
+    let ctx = StitchClient.init(th.testApp.client_app_id, { baseUrl: th.serverUrl });
+  	await ctx.then(stitchClient => {
```
edaniels: [style] tabbing is off here. Also, ctx should probably be renamed

src/client.js
```diff
@@ -87,6 +104,11 @@ export default class StitchClient {
     this.auth = new Auth(this, this.authUrl, authOptions);
     this.auth.handleRedirect();
     this.auth.handleCookie();
+
+    this._isInitialized = false;
+    this._willInitialize = this.auth._willInitialize.then(() => {
```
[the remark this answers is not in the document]
jsflax: `._willInitialize` seems like it should just be an exiting promise that can resolve always.

src/admin.js
```diff
@@ -1,14 +1,26 @@
...
-import StitchClient from './client';
+import { _StitchClient } from './client';
 import ADMIN_CLIENT_TYPE from './common';
 import { ADMIN_CLIENT_CODEC } from './auth/common';
 
 const v2 = 2;
 const v3 = 3;
 
-export default class Admin extends StitchClient {
+export class Admin {
```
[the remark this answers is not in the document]
jsflax: 👍

src/admin.js
```diff
@@ -1,14 +1,26 @@
...
+  constructor() {
+    throw new StitchError('Constructor undefined. Use `Admin.init` instead.');
```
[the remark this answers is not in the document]
jsflax: 👍

src/admin.js
```diff
@@ -1,14 +1,26 @@
...
+  }
+
+  static init(baseUrl) {
```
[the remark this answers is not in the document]
jsflax: 👍

src/admin.js
jsflax: Is this still applicable after our offline conversation?

src/admin.js
```diff
@@ -1,14 +1,26 @@
...
+    return new Promise((resolve, reject) => {
+      resolve(new _Admin(baseUrl));
+    });
+  }
+}
+
+export class _Admin extends _StitchClient {
```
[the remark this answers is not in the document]
jsflax: I don't think this is applicable after the offline conversation.

src/client.js
jsflax: 👍

src/client.js
jsflax: 👍

src/client.js
jsflax: 👍

src/client.js
jsflax: 👍

test/auth.test.js
jsflax: The test previously was directly write to storage with pre-massaged values. This probably wasn't correct in the first place, and using `storage` as it wasn't meant to be used isn't great in a test not even testing `storage`.

test/stitch_client.test.js
jsflax: 👍

src/admin.js
edaniels: Yes, this will need to call some "constructor" like method that can do async storage prep and then give you back a StitchClient (non-concrete since no interfaces here)

src/admin.js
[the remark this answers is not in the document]
edaniels: Correct. More specifically, there would be an `AdminClientFactory` and an `AdminClient` where the former is default exported and latter is not exported, but documented.

`import AdminClientFactory from "mongodb-stitch"`
`AdminClientFactory.create`

test/auth.test.js
edaniels: 👍

jsflax:
@edaniels PTAL

src/admin.js
```diff
@@ -1,14 +1,26 @@
...
+import { StitchClient } from './client';
 import ADMIN_CLIENT_TYPE from './common';
 import { ADMIN_CLIENT_CODEC } from './auth/common';
 
 const v2 = 2;
 const v3 = 3;
 
-export default class Admin extends StitchClient {
+export class StitchAdminFactory {
+  constructor() {
```
edaniels: [q] We don't even need this here right? They can totally make a factory but it'd be useless

src/admin.js
```diff
@@ -1,14 +1,26 @@
...
+  }
+
+  static create(baseUrl) {
+    return new Promise((resolve, reject) => {
+      resolve(new StitchAdminClient(baseUrl)._willInitialize);
+    });
+  }
+}
+
+export class StitchAdminClient extends StitchClient {
```
edaniels: [minor] We shouldn't be exporting the client. Only the `StitchAdminFactory` should be exported (can be default as well)

src/auth/index.js
```diff
@@ -11,27 +11,57 @@ const jwtDecode = require('jwt-decode');
...
+  }
+
+  static create(client, rootUrl, options) {
+    return new Promise((resolve, reject) => {
+      resolve(new Auth(client, rootUrl, options)._willInitialize);
+    });
+  }
+}
+
+export class Auth {
```
edaniels: [nit] I think this does not need to be exported either

src/auth/index.js
```diff
@@ -11,27 +11,57 @@ const jwtDecode = require('jwt-decode');
...
   constructor(client, rootUrl, options) {
-    let namespace;
-    if (!client || client.clientAppID === '') {
-      namespace = 'admin';
-    } else {
-      namespace = `client.${client.clientAppID}`;
-    }
+    this._willInitialize = new Promise((resolve, reject) => {
```
edaniels: [minor] I think we discussed this (at couches). I'd prefer if there were no way to construct any of these "concrete" classes. The `_willInitialize` concept across the two clients and auth feels kind of awkward that we do a `new` + `. _willInitialize `. What do you think of doing a separate function that is not exported that does `Object.create(Class.prototype)` and then taking the body of this here and appending to that?

Then you could write in the `create` functions: `resolve(createAuth())`

src/auth/index.js
```diff
@@ -11,27 +11,57 @@ const jwtDecode = require('jwt-decode');
...
+        this.storage.get(authCommon.REFRESH_TOKEN_KEY),
+        this.storage.get(authCommon.USER_LOGGED_IN_PT_KEY),
+        this.storage.get(authCommon.DEVICE_ID_KEY)
+      ]).then(([auth, rt, loggedInProviderType, deviceId]) => {
+        this.auth = auth;
+        this.authedId = auth.userId;
+        this.rt = rt;
+        this.loggedInProviderType = loggedInProviderType;
+        this.deviceId = deviceId;
+        this._isInitialized = true;
```
edaniels: [nit] I feel it cleaner to have no `_isInitialized` like variables as I mentioned in previous review. Instead, why not return the `Promise.all`? That way we get rejections to happen correctly.

src/auth/index.js
```diff
@@ -231,41 +263,41 @@ export default class Auth {
...
-      // do not include everything. This merging behavior is safe so long as json
-      // value responses with absent fields do not indicate that the field should
-      // be unset.
-      if (json[this.codec.accessToken]) {
-        newUserAuth.accessToken = json[this.codec.accessToken];
-      }
-      if (json[this.codec.userId]) {
-        newUserAuth.userId = json[this.codec.userId];
-      }
+    if (authType) {
```
edaniels: This feels... cleaner :)

src/client.js
```diff
@@ -23,70 +23,95 @@ const API_TYPE_CLIENT = 'client';
...
+
+  static create(clientAppID, options = {}) {
+    return new Promise((resolve, reject) => {
+      resolve(new StitchClient(clientAppID, options)._willInitialize);
+    });
+  }
+}
+
+/**
+ * Create a new _StitchClient instance.
```
edaniels: [nit] This comment may not apply anymore. We really do want the StitchClient documented, just not exported

src/client.js
```diff
@@ -23,70 +23,95 @@ const API_TYPE_CLIENT = 'client';
...
-      authOptions.storageType = options.storageType;
-    }
-    if (options.platform) {
-      authOptions.platform = options.platform;
-    }
-    if (options.authCodec) {
-      authOptions.codec = options.authCodec;
-    }
+      const authFactory = AuthFactory.create(this, this.authUrl, authOptions);
+      authFactory.then(auth => {
```
edaniels: With my suggestion, this would likely become just a return since we are outside of constructor.

jsflax:
@edaniels PTAL

src/auth/index.js
```diff
@@ -26,12 +37,33 @@ export default class Auth {
...
-    this.storage = createStorage(options);
-    this.providers = createProviders(this, options);
+    auth.client = client;
+    auth.rootUrl = rootUrl;
+    auth.codec = options.codec;
+    auth.platform = options.platform || _platform;
+    auth.storage = createStorage(options);
+    auth.providers = createProviders(auth, options);
+
+    Promise.all([
```
edaniels: From the previous round of comments but looks like it was not implemented: this should be what is returned instead of doing `return new Promise((resolve, reject) => {` and resolving with `auth`. You can do the `return Promise.all().then(... return auth)`

src/auth/index.js
```diff
@@ -231,41 +265,41 @@ export default class Auth {
...
+    // value responses with absent fields do not indicate that the field should
+    // be unset.
+    if (json[this.codec.accessToken]) {
+      newUserAuth.accessToken = json[this.codec.accessToken];
+    }
+    if (json[this.codec.userId]) {
+      newUserAuth.userId = json[this.codec.userId];
+    }
+
+    this.auth = Object.assign(!!this.auth && {}, newUserAuth);
```
edaniels: [nit] reduce the magic with `this.auth ? this.auth : {}`

src/client.js
```diff
@@ -84,9 +92,27 @@ export default class StitchClient {
       authOptions.codec = options.authCodec;
     }
 
-    this.auth = new Auth(this, this.authUrl, authOptions);
-    this.auth.handleRedirect();
-    this.auth.handleCookie();
+    const authFactory = AuthFactory.create(stitchClient, stitchClient.authUrl, authOptions);
+    authFactory.then(auth => {
+      stitchClient.auth = auth;
+      stitchClient.auth.handleRedirect();
```
edaniels: [major] These two return promises.
[minor] Same comment as before about returning a `Promise.all` instead of creating a new Promise at the beginning of this function. Last `.then` can return the stitchClient

src/client.js
```diff
@@ -84,9 +92,27 @@ export default class StitchClient {
...
+      stitchClient.auth.handleCookie();
+
+      resolve(stitchClient);
+    });
+  });
+}
+/**
+ * Create a new _StitchClient instance.
```
edaniels: This was also not addressed from last comment. This docstring is incorrect and we are exposing the client. It's just not constructible.

edaniels:
Reminder to update README

src/client.js
```diff
@@ -137,27 +163,19 @@ export default class StitchClient {
...
-        this.isAuthenticated(),
-        this.auth.getLoggedInProviderType()
-      ]
-    ).then(([isAuthenticated, loggedInProviderType]) => {
-      if (isAuthenticated) {
-        if (providerType === PROVIDER_TYPE_ANON && loggedInProviderType === PROVIDER_TYPE_ANON) {
-          return false; // is authenticated, skip log in
-        }
+    const authenticateFn = () =>
+      this.auth.provider(providerType).authenticate(options).then(() => this.auth.authedId);
```
edaniels: [nit] This should use this.authedId()

src/client.js
```diff
@@ -84,9 +92,27 @@ export default class StitchClient {
       authOptions.codec = options.authCodec;
     }
 
-    this.auth = new Auth(this, this.authUrl, authOptions);
-    this.auth.handleRedirect();
-    this.auth.handleCookie();
+    const authFactory = AuthFactory.create(stitchClient, stitchClient.authUrl, authOptions);
```
edaniels: [nit] This is really an authPromise

src/auth/index.js
edaniels: Actually, I'm pretty sure the code as is, has a bug and loses the past auth information rendering this.auth to always be set to newUserAuth, not merged. This then causes `this.authedId = this.auth.userId;` to be set to undefined on a token refresh and then permanently lost.

src/auth/index.js
```diff
@@ -26,12 +37,33 @@ export default class Auth {
...
+    auth.platform = options.platform || _platform;
+    auth.storage = createStorage(options);
+    auth.providers = createProviders(auth, options);
+
+    Promise.all([
+      auth._get(),
+      auth.storage.get(authCommon.REFRESH_TOKEN_KEY),
+      auth.storage.get(authCommon.USER_LOGGED_IN_PT_KEY),
+      auth.storage.get(authCommon.DEVICE_ID_KEY)
+    ]).then(([authObj, rt, loggedInProviderType, deviceId]) => {
```
edaniels: [q] If the clients rely on this resolving, are we assuming that a reject from storage always means there was a bad failure. In addition to that, does that mean we assume the contract is if you get something that isn't set as a key, it returned undefined/null? If yes to both of those, then this is fine as is; otherwise, we'd need to handle the reject case in the clients to continue on.

src/auth/index.js
jsflax: 👍

src/auth/index.js
jsflax: 👍

src/client.js
jsflax: 👍

src/client.js
jsflax: The docstring is wrong, but we discussed how we still need to expose the client internally.

src/client.js
jsflax: 👍

src/client.js
jsflax: 👍

src/auth/index.js
jsflax: So how do we handle the rejection case here then?

PR #111, opened 2018-01-31
STITCH-1046 JS SDK: Add ability for a user to link to any non-anonymous identity

jsflax:


src/admin.js
```diff
@@ -241,6 +241,10 @@ export class StitchAdminClient extends StitchClient {
             })
           }),
 
+          userRegistrations: () => ({
+            sendConfirm: (email) => api._post(`${appUrl}/user_registrations/by_email/${email}/send_confirm`)
```
edaniels: [nit] `sendConfirmationEmail`

src/auth/providers.js
```diff
@@ -11,6 +11,14 @@ export const PROVIDER_TYPE_GOOGLE = 'google';
 export const PROVIDER_TYPE_FACEBOOK = 'facebook';
 export const PROVIDER_TYPE_MONGODB_CLOUD = 'mongodbCloud';
 
+function urlWithParams(url, link) {
```
edaniels: [nit] This isn't really URL with params, it's URL with link aded on if true. How about `urlWithLinkParam`

src/auth/providers.js
```diff
@@ -217,15 +233,19 @@ function apiKeyProvider(auth) {
...
+    authenticate: (key, link) => {
       const deviceId = auth.getDeviceId();
       const device = auth.getDeviceInfo(deviceId, !!auth.client && auth.client.clientAppID);
       const fetchArgs = common.makeFetchArgs(
         'POST',
         JSON.stringify({ 'key': key, 'options': { device } })
       );
       fetchArgs.cors = true;
-      return fetch(`${auth.rootUrl}/${loginRoute}`, fetchArgs)
+      if (link) {
```
edaniels: [nit] Can we refactor this into another function like `fetchArgsWithLink`? Probably needs to be a method on auth though.

src/client.js
```diff
@@ -158,16 +162,20 @@ export class StitchClient {
...
     const authenticateFn = () =>
-      this.auth.provider(providerType).authenticate(options).then(() => this.authedId());
+      this.auth.provider(providerType).authenticate(options, link).then(() => this.authedId());
 
     if (this.isAuthenticated()) {
       if (providerType === PROVIDER_TYPE_ANON && this.auth.getLoggedInProviderType() === PROVIDER_TYPE_ANON) {
         return Promise.resolve(this.auth.authedId); // is authenticated, skip log in
       }
 
+      if (link) {
```
edaniels: [minor] I believe this belongs outside of the isAuthenticated block. The logic is, is this link? if yes, throw if not authenticated; otherwise call `authenticateFn`. If no, continue on with existing code. Can you add a test case that we throw when you try to link without being authed?

src/auth/providers.js
```diff
@@ -217,15 +233,19 @@ function apiKeyProvider(auth) {
      * @param {String} key the key for authentication
      * @returns {Promise} a promise that resolves when authentication succeeds.
      */
-    authenticate: key => {
+    authenticate: (key, link) => {
```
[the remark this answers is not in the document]
edaniels: The server will error. We could throw but I think this is okay for now

test/auth.test.js
```diff
@@ -76,6 +76,46 @@ describe('Auth', () => {
   });
 });
 
+describe('Auth linking', () => {
```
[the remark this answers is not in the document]
edaniels: I'd say it's just duplicating work for something the server already exhaustively tests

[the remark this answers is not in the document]
jsflax: This test is just making sure that the flow works.

src/client.js
```diff
@@ -149,6 +149,10 @@ export class StitchClient {
     return this.auth.provider('userpass').register(email, password, options);
   }
 
+  linkWithProvider(providerType, options = {}) {
```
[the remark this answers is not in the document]
jsflax: 👍

jsflax:
@edaniels @adamchel PTAL. All notes were followed; I did completely separate linking and authenticating, which satisfies the requirements. Also added a test for the `isAuthenticated()` check in the linking method.

src/client.js
```diff
@@ -149,6 +149,22 @@ export class StitchClient {
     return this.auth.provider('userpass').register(email, password, options);
   }
 
+
+  /**
+   * Links the currently logged in account with another existing account.
```
adamchel: [minor] I think it's a little misleading to say link the current account with another existing account since it can't actually link two existing users. Maybe "Links the currently logged in user with another identity", and for the `providerType` param: "the authentication provider of the identity to link with this user (e.g 'userpass', 'facebook', google')"

PR #113, opened 2018-02-05
STITCH-1166 Revamp SDK documentation

adamchel:
In the process of going through our docstrings to hide things that don't need to be in the public docs, I decided it might be worthwhile to explore a new documentation framework since the nav on our current one could be better. I ended up going with https://github.com/documentationjs/documentation since it supports the JSDoc syntax which we've been using, as well as Flow type annotations, should we go that route.

I've staged the updated documentation at https://adamchel.github.io/stitch-js-sdk/

Here's a summary of the changes:
- in `package.json`, I updated the `docs` command to use the new framework, and I added the necessary dependencies as devDependencies. After installing the dependencies, `yarn docs` should generate the docs and put them in the `docs` directory.
- I added a new `docs_overview.md` that serves as the landing content for the new SDK docs. I did this instead of including the `README.md` since the README contains some things that wouldn't be immediately relevant to an SDK user and would make more sense in the README for the repo.
- I copied the default theme from the documentationjs/documentation repo into the `docs_theme` directory and modified it significantly to make a more visually appealing theme than the default.
- I went through the existing doc strings and tagged ones we don't want visible with `@private`, and I updated some of the content so it makes more sense with the way things are now laid out.


docs_overview.md
```diff
@@ -0,0 +1,73 @@
...
+
+[MongoDB Stitch Announcements - Google Group](https://groups.google.com/d/forum/mongodb-stitch-announce)
+
+[MongoDB Stitch Documentation](https://docs.mongodb.com/stitch/)
+
+[JavaScript SDK GitHub Repository](https://github.com/mongodb/stitch-js-sdk)
+
+### Basic Usage
+
+#### Construct a simple app-wide client
```
edaniels: It'd be nice if there was a way to source this from the README or some single location in the event we need to make updates.

src/auth/providers.js
```diff
@@ -353,7 +375,10 @@ function facebookProvider(auth) {
   };
 }
 
-/** @namespace */
+/**
```
edaniels: [q] Can we somehow get the list of providers in https://adamchel.github.io/stitch-js-sdk/#stitchclient's authenticate method?

src/client.js
```diff
@@ -23,13 +23,26 @@ const API_TYPE_CLIENT = 'client';
...
 /**
-  * Factory class to create a new StitchClient asynchronously.
+  * StitchClientFactory is a singleton factory class which can be used to
+  * asynchronously create instances of {@link StitchClient}. StitchClientFactory
+  * is not meant to be instantiated. Use the static `create()` method to build
+  * a new StitchClient.
   */
 export class StitchClientFactory {
+  /**
+   * @hideconstructor
```
edaniels: 😎

src/client.js
```diff
@@ -192,7 +207,7 @@ export class StitchClient {
   }
 
   /**
-   * Ends the session for the current user.
+   * Ends the session for the current user, and clears auth information from local storage.
```
edaniels: [nit] change local storage to storage since it's not always local

docs_overview.md
adamchel: It seems like we wouldn't be able to do this and have both our Github README and Overview have the format we want: https://github.com/github/markup/issues/172#issuecomment-33241601

src/auth/providers.js
adamchel: Just listed them in the docstring

PR #118, opened 2018-02-21
STITCH-1217 - Handle irrelevant hash fragment correctly

edaniels:
@jsflax this code causes some nasty issues on sites using fragments. They would essentially fail every single time a page is hit with a fragment since `redirectProvider` is undefined. It was incorrect to fail if `!redirectProvider `. The proper condition is redirect state found and `!redirectProvider`.

This code also fixes the returning early which worked in synchronous code but not here since we move onto the next promise that has undefined arguments. Now we reject to the last catch and continue on happily provided no real error was thrown/rejected.

### DIFF
--- a/src/auth/index.js
+++ b/src/auth/index.js
@@ -126,24 +126,22 @@ export class Auth {
       return;
     }
 
-    let redirectProvider;
     return Promise.all([
       this.storage.get(authCommon.STATE_KEY),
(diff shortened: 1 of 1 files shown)


src/auth/index.js
```diff
@@ -126,33 +126,31 @@ export class Auth {
       return;
     }
 
-    let redirectProvider;
```
edaniels: just an idiomatic clean up here

PR #120, opened 2018-03-02
STITCH-1223 - Add app export api to admin client

makesitgo:


jsflax:
@makesitgo Is it too soon to write a test for this?

makesitgo:
@jsflax ... testing provides minimal benefit at the moment (it would amount to asserting `content-type` and `content-disposition` headers).  this is mostly a pass through http response to the UI so the browser can save the target file.

edaniels:
Let's test it later, said the wise man.

PR #128, opened 2018-05-16
STITCH-1510 JS SDK: Add encoding/decoding to all extJson calls

jsflax:


jsflax:
@edaniels @adamchel Hold off, this is not ready.

jsflax:
@adamchel @edaniels PTAL

core/src/lib/internal/CoreStitchAppClient.ts
```diff
@@ -16,10 +17,9 @@ export default class CoreStitchAppClient {
   }
 
   public callFunctionInternal<T>(name: string, ...args: any[]): Promise<T> {
-    // TODO: STITCH-1510
-    return this.authRequestClient.doAuthenticatedJSONRequest(
+    return EJSON.parse(this.authRequestClient.doAuthenticatedJSONRequest(
```
adamchel: I don't think this will work. `doAuthenticatedJSONRequest()` returns a `Promise<Response>` so I think you'll need to get the result of the promise and call EJSON.parse on the response body.

adamchel: Also I think this logic also needs to go in `CoreStitchService`

jsflax: I agree about point 1, but point 2 I'm not sure about. Seems like not everything from `CoreStitchService` may be extJson.

jsflax:
I'll write some tests– decoding should work as is (once I fix the promise part): https://github.com/mongodb-js/mongodb-extjson#deserialize-a-document

core/src/lib/internal/CoreStitchAppClient.ts
adamchel: just `doAuthenticatedJSONRequest`? Are there any endpoints that don't accept a JSON body, but still return extended JSON?

edaniels: There are not in the client API. All functions require a JSON body.

jsflax:
@edaniels @adamchel PTAL

jsflax:
@adamchel I agree, though it should be a bit more indirect. We will have to rewrite most of these tests once we add sinon anyway, so I figure we will wait for now.

PR #131, opened 2018-05-22
STITCH-1548 Implement User API Key client

adamchel:
This is mostly the same as Java/Swift, but I used a string instead of an object ID in the UserAPIKey model since AFAIK we don’t have BSON in the core module, and I didn’t want to bring it in just for the sake of having an ObjectId.

It would be straightforward for me to make it an ObjectId use https://www.npmjs.com/package/bson-objectid, so let me know if we think we should do that at the very least.

I didn’t add any tests since there are currently no tests in the web module and I’m assuming Jason will be adding some in his sinon work. I know this isn’t ideal but as we’ve said in other PRs, I don’t want to spend time writing tests that are going to be rewritten soon. It seems like once Jason is done with iOS stuff we’ll both be able to jump on the Sinon/Cuckoo work.

**Drive-bys**:
- Clean up unnecessary auth provider clients.
    - Reordered the arguments of credential constructors so users can omit the provider name argument when constructing credentials directly.
- Update user/password auth provider client to properly deal with promises
- Remove `@link` references in docstrings (they’re not conventional for JS docs and they break VSCode tooltips)
- Update some stuff in StitchAuthImpl to get it compiling


web/src/lib/core/auth/providers/internal/AuthProviderClientSupplier.ts
```diff
@@ -1,9 +1,10 @@
-import { StitchAuthRoutes, StitchRequestClient } from "stitch-core";
+import { StitchAuthRoutes, StitchRequestClient, StitchAuthRequestClient } from "stitch-core";
 
-interface AuthProviderClientSupplier<T> {
+interface AuthProviderClientSupplier<ClientT> {
```
edaniels: [nit] RequestClientT

core/src/lib/internal/common/Assertions.ts
```diff
@@ -0,0 +1,13 @@
+class Assertions {
```
edaniels: nice

core/src/lib/auth/providers/anonymous/AnonymousCredential.ts
```diff
@@ -24,7 +24,7 @@ export default class AnonymousCredential implements StitchCredential {
    */
   public readonly providerCapabilities = new ProviderCapabilities(true);
 
-  constructor(providerName: string) {
+  constructor(providerName: string = ProviderTypes.ANON) {
```
edaniels: [minor] This should really be changed to match java where we have a DEFAULT_NAME and TYPE value inside the *AuthProvider file itself. The type is technically not a name.

web/src/lib/core/auth/providers/internal/AuthProviderClientSupplier.ts
adamchel: This is actually not the request client type, but the type of Auth provider client that it returns, so I believe it's correct as is

core/src/lib/internal/common/Assertions.ts
```diff
@@ -0,0 +1,13 @@
...
+  /**
+   * Throw IllegalStateException if key is not present in map.
```
adamchel: oops, I neglected to update this docstring, I'll fix that in the next commit

web/src/lib/core/auth/providers/internal/AuthProviderClientSupplier.ts
edaniels: Yep. I thought this was the provider client signature. My bad

core/src/lib/StitchAppClientConfiguration.ts
```diff
@@ -1,9 +1,37 @@
...
+      config.storage,
+      config.dataDirectory,
+      config.transport
+    );
+    this.clientAppId = clientAppId;
+    this.localAppVersion = localAppVersion;
+    this.localAppName = localAppName;
+  }
+
+  public builder(): StitchAppClientConfiguration.Builder {
```
adamchel: The main reason I had to declare the inner classes in this way is so that these `builder()` functions could return the proper type, which they were previously not doing. Once I gave these functions return types, the TS compiler caught a lot of issues.

core/src/lib/StitchAppClientConfiguration.ts
```diff
@@ -16,7 +44,7 @@ export default class StitchAppClientConfiguration extends StitchClientConfigurat
     public localAppName: string;
     public localAppVersion: string;
 
-    private constructor(config?: StitchAppClientConfiguration) {
+    public constructor(config?: StitchAppClientConfiguration) {
```
adamchel: The main disadvantage of this workaround for inner classes is that the inner class constructor can no longer be private because to the outer class, the inner class declaration is at the top-level and not within the class.

core/src/lib/auth/internal/CoreStitchAuth.ts
```diff
@@ -159,11 +154,10 @@ export default abstract class CoreStitchAuth<TStitchUser extends CoreStitchUser>
     stitchReq: StitchAuthDocRequest
   ): Promise<Response> {
     const newReqBuilder = stitchReq.builder;
-    newReqBuilder.withBody(stitchReq.document);
-    const newHeaders = newReqBuilder.headers; // This is not a copy
+    newReqBuilder.withBody(JSON.stringify(stitchReq.document));
+    const newHeaders = newReqBuilder.headers || {}; // This is not a copy
```
adamchel: For example, now that `newReqBuilder` is typed, the TS compiler was able to catch the fact that the `newReqBuilder.headers` could be undefined, so I added this safeguard in a lot of places

core/src/lib/internal/CoreStitchAppClient.ts
```diff
@@ -15,19 +15,19 @@ export default class CoreStitchAppClient {
...
+  public callFunctionInternal<T>(name: string, args: any[]): Promise<T> {
     return this.authRequestClient.doAuthenticatedJSONRequest(
       this.getCallFunctionRequest(name, args)
     );
   }
 
   private getCallFunctionRequest(
     name: string,
-    ...args: any[]
+    args: any[]
```
adamchel: Got rid of variadic/rest parameters here, since it was triply nesting the variadic args into something that looked like `[[["the", "actual", "arguments"]]]`. The interface in the web module is still exposed as variadic/rest parameters, but I'm open to making them non-variadic for the sake of consistency with the other SDKs. Would like your thoughts.

core/src/lib/internal/net/BasicRequest.ts
```diff
@@ -1,11 +1,30 @@
 import Method from "./Method";
 
-export default class BasicRequest {
-  public static Builder = class {
+export class BasicRequest {
+  public readonly method: Method;
+  public readonly url: string;
+  public readonly headers: { [key: string]: string };
+  public readonly body?: string;
```
adamchel: Per the discussion in Slack earlier, I made request bodies and response bodies strings instead of `any`s so the typescript compiler actually does type checking for us, and it was able to catch a bunch of issues where we weren't decoding or encoding JSON into or from a string in several places.

core/src/lib/internal/net/FetchTransport.ts
```diff
@@ -1,23 +1,31 @@
-import BasicRequest from "./BasicRequest";
+import { BasicRequest } from "./BasicRequest";
 import Response from "./Response";
 import Transport from "./Transport";
+require('es6-promise').polyfill();
```
adamchel: This is how you're supposed to use `fetch-everywhere` according to its docs. This seems weird when mixed with the `import`s but I wasn't sure if there's another proper way to do it in ES6.

edaniels: Oh well!

web/src/lib/core/auth/providers/userpassword/UserPasswordAuthProviderClient.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  /**                                                                                                                                                                                                               
+   * Sends a user a password reset email for the given email.                                                                                                                                                       
+   *                                                                                                                                                                                                                
+   * @return A Promise that completes when the reqest request completes/fails.                                                                                                                                 
+   */
+  sendResetPasswordEmail(email: string): Promise<void>
+}
+
+export namespace UserPasswordAuthProviderClient {
```
edaniels: 8)

PR #133, opened 2018-05-31
STITCH-1565 Web SDK: Move to a mocking framework

jsflax:
Note: The evergreen file is in a partial state until integration tests have been added.


core/src/lib/services/internal/CoreStitchServiceImpl.ts
```diff
@@ -0,0 +1,47 @@
+import {
+  CoreStitchService,
+  StitchAuthRequestClient,
+  StitchServiceRoutes
+} from "../..";
+import { StitchAuthDocRequest } from "../../internal/net/StitchAuthDocRequest";
+import Method from "../../internal/net/Method";
+
+export default class CoreStitchServiceImpl implements CoreStitchService {
```
edaniels: [nit] Rename `CoreStitchService` and `StitchService` to `CoreStitchServiceClient` and `StitchServiceClient`

core/src/lib/internal/net/StitchAuthDocRequest.ts
```diff
@@ -20,8 +23,13 @@ export namespace StitchAuthDocRequest {
     public document: object;
     public useRefreshToken: boolean;
 
-    public constructor(request?: StitchAuthRequest) {
+    public constructor(request?: StitchAuthDocRequest) {
       super(request);
+
+      if (request !== undefined) {
```
edaniels: [q] when would this be undefined? Should the request just be non optional

core/src/lib/auth/providers/userapikey/CoreUserAPIKeyAuthProviderClient.ts
```diff
@@ -19,150 +19,149 @@ enum ApiKeyFields {
...
+    reqBuilder.withMethod(Method.GET).withPath(this.baseRoute);
+    reqBuilder.withRefreshToken();
+
     return this.requestClient
       .doAuthenticatedRequest(reqBuilder.build())
       .then(response => {
-        if(Array.isArray(response.body)) {
-          const keys = Array.from(response.body)
-          return keys.map(value => UserAPIKey.readFromAPI(value))
+        const json = JSON.parse(response.body!);
```
edaniels: [minor] This should ideally be using the `doAuthenticatedRequest` with the simple `Decoder<T>` we discussed.

edaniels: Same thing applies to other PRs here.

core/src/lib/StitchClientErrorCode.ts
```diff
@@ -3,10 +3,10 @@
...
-    LOGGED_OUT_DURING_REQUEST,
-    MISSING_URL,
-    MUST_AUTHENTICATE_FIRST,
-    USER_NO_LONGER_VALID,
-    COULD_NOT_LOAD_PERSISTED_AUTH_INFO,
-    COULD_NOT_PERSIST_AUTH_INFO
+  LOGGED_OUT_DURING_REQUEST,
+  MISSING_URL,
+  MUST_AUTHENTICATE_FIRST,
+  USER_NO_LONGER_VALID,
```
edaniels: make like other enums

core/src/lib/auth/internal/CoreStitchAuth.ts
```diff
@@ -171,29 +161,31 @@ export default abstract class CoreStitchAuth<TStitchUser extends CoreStitchUser>
...
   }
 
-  protected abstract onAuthEvent();
-
   /**
    * Authenticates the `CoreStitchAuth` using the provided `StitchCredential`. Blocks the current thread until the
    * request is completed.
    */
-  protected loginWithCredential(
+  public loginWithCredential(
```
edaniels: suffix internal

core/src/lib/internal/net/StitchAuthRequest.ts
```diff
@@ -37,8 +39,17 @@ export namespace StitchAuthRequest {
...
     }
 
+    public withShouldRefreshOnFailure(shouldRefreshOnFailure: boolean): this {
+      this.shouldRefreshOnFailure = shouldRefreshOnFailure;
+      return this;
+    }
+
     public build(): StitchAuthRequest {
-      return new StitchAuthRequest(super.build(), this.useRefreshToken);
+      return new StitchAuthRequest(
```
edaniels: Add
```
if (useRefreshToken) {
        shouldRefreshOnFailure = false;
      }
```

jsflax:
@edaniels PTAL

core/src/__tests__/internal/net/StitchRequestClientUnitTests.ts
```diff
@@ -0,0 +1,300 @@
...
+      body: '{"error": "bad", "error_code": "InvalidSession"}',
+      headers,
+      statusCode: 502
+    });
+
+    expect(stitchRequestClient.doRequest(builder.build())).rejects.toEqual(
+      new StitchServiceException("bad", StitchServiceErrorCode.InvalidSession)
+    );
+
+    headers[Headers.CONTENT_TYPE_CANON] = ContentTypes.APPLICATION_JSON;
```
adamchel: I think this test is incorrect (the Java one is incorrect too). this should be something like `const canonicalHeaders = headers = { [Headers.CONTENT_TYPE]: ContentTypes.APPLICATION_JSON };`, and you should have the server return the `canonicalHeaders` for the next request. The way it's written right now, the headers contain both a `Content-Type` and a `content-type`, and the error handling code checks for the lowercase version.

As the code stands right now, I think the test would fail. You need to update `Response` so that when it is created, the headers are processed by having the keys be lowercased. See the main constructor in `Response.java` in the Java SDK.

core/src/lib/StitchError.ts
```diff
@@ -70,8 +72,10 @@ export default class StitchError {
    */
   private static handleRichError(response: Response, body: string): string {
     if (
-      response.headers[Headers.CONTENT_TYPE] !== undefined ||
-      response.headers[Headers.CONTENT_TYPE] !== ContentTypes.APPLICATION_JSON
+      response.headers[Headers.CONTENT_TYPE] === undefined ||
```
adamchel: Regarding my other comment, this is checking for the all lowercase version of the header (which is correct here), the headers from the response need to be preprocessed for the rich error to be properly decoded.

core/src/lib/auth/internal/CoreStitchAuth.ts
```diff
@@ -137,28 +141,23 @@ export default abstract class CoreStitchAuth<TStitchUser extends CoreStitchUser>
...
    */
-  public doAuthenticatedJSONRequest(
-    stitchReq: StitchAuthDocRequest
-  ): Promise<any> {
-    return this.doAuthenticatedJSONRequestRaw(stitchReq)
-      .then(response => EJSON.parse(response.body, {strict: false}))
-      .catch(err => { throw StitchError.wrapDecodingError(err) })
-  }
+  public doAuthenticatedJSONRequest<T>(
+    stitchReq: StitchAuthDocRequest,
```
adamchel: [minor] AFAIK now that we properly encode bodies at request build time there's no more reason for this to be a `StitchAuthDocRequest`, it can just be a `StitchAuthRequest`, and you can rename it to `doAuthenticatedRequest<T>`

core/src/lib/auth/internal/StitchAuthRequestClient.ts
```diff
@@ -21,15 +21,4 @@ interface StitchAuthRequestClient {
    * - returns: An `Any` representing the response body as decoded JSON.
    */
   doAuthenticatedJSONRequest(stitchReq: StitchAuthDocRequest): Promise<any>;
```
adamchel: [nit] regarding my other comment, `StitchAuthDocRequest` -> `StitchAuthRequest`

core/src/lib/auth/providers/userapikey/CoreUserAPIKeyAuthProviderClient.ts
```diff
@@ -19,150 +19,149 @@ enum ApiKeyFields {
 /**
  * A client for the user API key authentication provider which can be used to obtain a credential for logging in.
  */
-export default abstract class CoreUserAPIKeyAuthProviderClient extends CoreAuthProviderClient<StitchAuthRequestClient>{
+export default abstract class CoreUserAPIKeyAuthProviderClient extends CoreAuthProviderClient<
+  StitchAuthRequestClient
```
adamchel: [nit] this looks really strange. was this the prettier? It does not look pretty haha. can you just line break before the `extends` and keep the generic information all on one line

core/src/lib/auth/providers/userpass/CoreUserPasswordAuthProviderClient.ts
```diff
@@ -18,17 +18,22 @@ enum ActionFields {
   TOKEN_ID = "tokenId"
 }
 
-export default abstract class CoreUserPasswordAuthProviderClient extends CoreAuthProviderClient<StitchRequestClient> {
-  protected constructor(
+export default class CoreUserPasswordAuthProviderClient extends CoreAuthProviderClient<
+  StitchRequestClient
```
adamchel: [nit] same prettier issue

core/src/lib/internal/net/Response.ts
```diff
@@ -1,7 +1,7 @@
 interface Response {
```
adamchel: re: my other comments about the header stuff, this should probably be a class with a constructor that preprocesses the headers to make them all have lowercased keys.

core/tsconfig.json
```diff
@@ -9,6 +9,7 @@
         "removeComments": true
     },
     "exclude": [
-        "node_modules"
+        "node_modules",
+        "./dist"
```
adamchel: [nice]

web/src/lib/core/Stitch.ts
```diff
@@ -5,96 +5,112 @@ import StitchAppClient from "./StitchAppClient";
...
     }
-  
-    public static get defaultAppClient(): StitchAppClient {
-      Stitch.ensureInitialized();
-      if (Stitch.defaultClientAppId === undefined) {
-        throw new Error("default app client has not yet been initialized/set");
-      }
-      return appClients[Stitch.defaultClientAppId];
+
+    // TODO: get app info
```
adamchel: [nit] is there a ticket for this? if so can you add the number here so we don't forget about it?

web/src/lib/core/auth/providers/userapikey/UserAPIKeyAuthProviderClient.ts
```diff
@@ -1,57 +1,65 @@
...
    */
-  disableApiKey(keyId: ObjectID): Promise<void>
+  disableApiKey(keyId: ObjectID): Promise<void>;
 }
 
 export namespace UserAPIKeyAuthProviderClient {
-  export const Factory: AuthProviderClientFactory<UserAPIKeyAuthProviderClient> =
-  new class implements AuthProviderClientFactory<UserAPIKeyAuthProviderClient> {
+  export const Factory: AuthProviderClientFactory<
+    UserAPIKeyAuthProviderClient
```
adamchel: [nit] another weirdly formatted generic

web/src/lib/core/auth/providers/userpassword/UserPasswordAuthProviderClient.ts
```diff
@@ -1,53 +1,63 @@
...
-    new class implements AuthProviderClientFactory<UserPasswordAuthProviderClient> {
-      public getClient(
-        authRequestClient: StitchAuthRequestClient, // this arg is ignored
-        requestClient: StitchRequestClient,
-        routes: StitchAuthRoutes): UserPasswordAuthProviderClient {
-        return new UserPasswordAuthProviderClientImpl(requestClient, routes);
-      }
-    }();
+  export const Factory: AuthProviderClientFactory<
+    UserPasswordAuthProviderClient
```
adamchel: [nit] weird generic formatting

web/src/lib/services/StitchService.ts
```diff
@@ -1,5 +1,5 @@
 interface StitchService {
-    callFunction(name: string, ...args: any[]): Promise<any>;
+  callFunction(name: string, ...args: any[]): Promise<any>;
```
adamchel: [nit q] Are we sure we want `args` to be variadic? If an interface above this ever exposes a variadic arg parameter to be passed as a list of args here,  it could end up being doubly nested here.

core/src/__tests__/auth/providers/userapikey/CoreUserApiKeyAuthProviderClientUnitTests.ts
```diff
@@ -0,0 +1,258 @@
...
+  keyToFetch?: string
+): Promise<any> {
+  const clientAppId = "my_app-12345";
+
+  const requestClientMock = mock(CoreStitchAuth);
+  const requestClient = instance(requestClientMock);
+
+  const routes = new StitchAppRoutes(clientAppId).authRoutes;
+
+  const client = new class extends CoreUserAPIKeyAuthProviderClient {
```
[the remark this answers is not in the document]
jsflax: This class used to be `abstract` on all of the SDKs that support abstractness. I can change it back.

In general, I wouldn't consider `abstract` to be in the context of "convention". It's a modifier that prevents instantiation, which I'd argue is a step above convention.

Anyway, I'll change this.

web/src/lib/core/auth/providers/userapikey/UserAPIKeyAuthProviderClient.ts
jsflax: This may be due to a character limit on the line. I'll look into it, but I'm not going to spend that much time for it. It's not _that_ ugly

web/src/lib/services/StitchService.ts
jsflax: That's true for any variadic parameter. We can change it to an array if we want, as that might be more comfortable for JS users, however, in some languages, variadic makes life a lot easier, and it's up to the dev to not make the double nesting mistake– not us.

core/src/__tests__/internal/net/StitchRequestClientUnitTests.ts
adamchel: Looks like the preprocessing is implemented now, but the mocked request client in the test is still returning both headers. 

You can change this line to create `const canonicalHeaders = headers = { [Headers.CONTENT_TYPE]: ContentTypes.APPLICATION_JSON };`, and pass `canonicalHeaders` in line 292 and the test will be correct, and with your new headers code I think it should pass.

web/src/lib/services/StitchService.ts
adamchel: I think an array is the right move, this isn't going to be the interface that people use right? From what I understand our services will wrap one of these.

adamchel: Anyway, you made the change and I think that's fine

jsflax: Yeah I'd say it's the right move for JS

jsflax:
@adamchel I'm going to merge this and make those changes in the next PR. It's odd already that we're squashing together all these tickets, and the code does work. Parity with Java will have to occur later. It's almost there.

PR #135, opened 2018-06-04
STITCH-1600 Web SDK: Implement integration tests

jsflax:
Main changes:
- `lerna` has been added: https://github.com/lerna/lerna . `lerna` will allow us to do simple multi-module support.
- The build process has changed due to `lerna`. This will be updating in contrib.
- `core-testutils`, `web-testutils`, and `web-coretest` are the big additions here. Many changes have been made to `admin` as well, to facilitate the integration tests.
- Many linter and prettier fixes.


packages/core/package.json
```diff
@@ -5,29 +5,19 @@
   "types": "dist/lib/index.d.ts",
   "description": "stitch",
   "dependencies": {
+    "bson": "^2.0.7",
+    "@types/bson": "^1.0.8",
```
adamchel: [minor] I don't think the `@types` should be dependencies of core (especially the jest and ones). This would seem to mean that all projects that include the `stitch-web` module would have to bring in these dependencies, even if they don't have Typescript.

If anything, these should just be made `devDependencies` of the module above that is trying to use them.

packages/core/src/lib/internal/net/FetchTransport.ts
```diff
@@ -21,13 +22,9 @@ export default class FetchTransport implements Transport {
       const body = values[1];
       const headers: { [key: string]: string } = {};
       response.headers.forEach((key, value) => {
-        headers[key] = value;
+        headers[value] = key;
```
adamchel: [nit] even if this is right, this appears incorrect. If the headers in the fetch response are actually reversed, change the foreach to be 
```
response.headers.forEach((value, key) => {
    headers[key] = value
```

packages/core/src/lib/internal/net/Response.ts
```diff
@@ -7,8 +7,9 @@ export default class Response {
     public readonly body?: string
   ) {
     // preprocess headers
-    Object.keys(headers).map(
-      (key, index) => (this.headers[key.toLocaleLowerCase()] = headers[key])
-    );
+
```
adamchel: [nit] extra whitespace

packages/core/src/lib/internal/net/StitchAuthDocRequest.ts
```diff
@@ -6,11 +6,13 @@ import { StitchAuthRequest } from "./StitchAuthRequest";
 import { StitchRequest } from "./StitchRequest";
 
 export class StitchAuthDocRequest extends StitchAuthRequest {
-  public readonly document: object;
-
-  public constructor(request: StitchRequest, document: object) {
-    super(request, false);
-    this.document = document;
+  public constructor(
```
adamchel: [minor] you should not be doing anything with `shouldRefreshToken` or `shouldRefreshOnFailure` in `StitchAuthDocRequest`. The superclass should be dealing with this. See the Java and Swift implementations.


The constructor should only accept a `StitchAuthRequest` and a `document`, calling the `StitchAuthRequest` constructor that accepts a `StitchAuthRequest`. I know that javascript doesn't let you have multiple constructors in a class, but there must be a way to do this without breaking the usefulness of inheritance.

packages/example/package.json
```diff
@@ -1,15 +1,12 @@
...
     "es6-promise": "^4.2.4",
     "fetch-everywhere": "^1.0.5",
-    "stitch-web": "file:../web"
-  },
-  "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "stitch-web": "^4.0.0"
   },
   "author": "",
   "license": "ISC"
```
adamchel: [nit] MongoDB author and Apache 2 license

packages/web-coretest/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "stitch-web-coretest",
+  "version": "4.0.0",
+  "description": "",
+  "main": "index.js",
+  "author": "",
+  "license": "ISC",
```
adamchel: [nit] mongodb author and apache 2 license

packages/web-testutils/package.json
```diff
@@ -1,14 +1,16 @@
...
-  "name": "stitch-admin",
-  "version": "1.0.0",
+  "name": "stitch-web-testutils",
+  "version": "4.0.0",
   "description": "",
-  "main": "index.js",
+  "main": "dist/BaseStitchIntTestWebHarness.js",
+  "types": "dist/BaseStitchIntTestWebHarness.d.ts",
   "author": "",
   "license": "ISC",
```
adamchel: [nit] MongoDB author and Apache 2 license

packages/core-testutils/src/lib/BaseStitchIntTest.ts
```diff
@@ -0,0 +1,118 @@
...
+  RuleCreator,
+  RuleResponse,
+  Service,
+  ServiceConfig,
+  ServiceResponse,
+  StitchAdminClient
+} from "stitch-admin";
+import { UserAPIKeyAuthProvider, UserPasswordCredential } from "stitch-core";
+
+export default abstract class BaseStitchIntTestHarness {
```
adamchel: [nit] rename this file to `BaseStitchIntTestHarness.ts`

packages/web/src/lib/core/Stitch.ts
```diff
@@ -8,21 +8,12 @@ const TAG = "Stitch";
...
-      );
-    }
-  }
-
   public static initialize() {
     if (Stitch.initialized) {
       return;
     }
 
     // TODO: get app info
```
adamchel: [nit] ticket?

packages/web/src/lib/core/StitchAppClient.ts
```diff
@@ -12,7 +12,7 @@ interface StitchAppClient {
 
   getServiceClient<T>(provider: ServiceClientProvider<T>): T;
 
-  callFunction(name: string, ...args: any[]): Promise<any>;
+  callFunction(name: string, args: any[]): Promise<any>;
```
adamchel: [note] let's make a note of this for when we play around with usability next week. Since this is not internal, I think this could go either way here for variadic vs. non-variadic.

packages/web/src/lib/core/internal/common/LocalStorage.ts
```diff
@@ -0,0 +1,124 @@
...
+  public set(key: string, value: string): any {
+    return localStorage.setItem(`${this.suiteName}.${key}`, value);
+  }
+  public remove(key: string): any {
+    return localStorage.removeItem(`${this.suiteName}.${key}`);
+  }
+}
+
+if (!window.localStorage) {
+  Object.defineProperty(
```
adamchel: [q] was this code you wrote? If not we should properly attribute it in a comment. or if you wrote the code then comment with a link to whatever specification you followed?

.evg.yml
```diff
@@ -65,8 +65,9 @@ functions:
           export STITCH_PATH=$GOPATH/src/github.com/10gen/stitch
           export PATH="$PATH:$STITCH_PATH/etc/transpiler/bin"
           which transpiler
-          cd stitch-js-sdk/test/bin
+          # cd stitch-js-sdk/test/bin
           echo "running stitch"
+          go run $STITCH_PATH/cmd/auth/user.go addUser -domainID 000000000000000000000000 -mongoURI mongodb://localhost:26000 -salt 'DQOWene1723baqD!_@#' -id "unique_user@domain.com" -password "password"
```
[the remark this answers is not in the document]
jsflax: It comes from the stitch readme. If this shouldn't be exposed, we need to change iOS as well.

packages/web/src/lib/core/StitchAppClient.ts
jsflax: To be honest, I ran into the exact issue we were talking about– doubling up the use of an array. I'm not sure variadic works as well in a language as loose as TS.

packages/web/src/lib/core/internal/common/LocalStorage.ts
jsflax: I altered the code from Mozilla: https://developer.mozilla.org/en-US/docs/Web/API/Storage/LocalStorage. I'll add a comment.

jsflax:
@adamchel PTAL

PR #136, opened 2018-06-05
STITCH-1471 Web SDK: Implement Twilio Service in Core and Web

jsflax:


packages/core/src/lib/internal/net/StitchAuthDocRequest.ts
```diff
@@ -54,7 +54,7 @@ export namespace StitchAuthDocRequest {
       if (this.headers === undefined) {
         this.withHeaders({});
       }
-      
+
```
adamchel: [nit] add back the application/json content-type header

PR #137, opened 2018-06-05
STITCH-1468 Web SDK: Implement SES Service in Core and Web

jsflax:


packages/core/src/lib/auth/internal/StitchAuthRequestClient.ts
```diff
@@ -19,5 +20,8 @@ export default interface StitchAuthRequestClient {
    *
    * - returns: An `Any` representing the response body as decoded JSON.
    */
-  doAuthenticatedJSONRequest<T>(stitchReq: StitchAuthRequest): Promise<T>;
+  doAuthenticatedJSONRequest<T>(
```
adamchel: [nit] this could be a good opportunity to rename this function to `doAuthenticatedRequest()` for the sake of parity with iOS and Java, or if it gets complicated with TypeScript and overloading, maybe name it something more descriptive like `doAuthenticatedRequestWithDecoding`

PR #143, opened 2018-06-18
STITCH-1683 Add endpoint for resuming suspended events

marcgurney:


edaniels:
This will need to go off into a branch I can create for 3.x

edaniels:
It's called `support/3.x`

PR #144, opened 2018-06-20
STITCH-1675 STITCH-1558 - Prepare 4.0 SDK for publishing

edaniels:
lots o drivebys
i'll see what I can mark


packages/browser/core/rollup.config.js
```diff
@@ -0,0 +1,110 @@
...
+        'Map',
+        'MaxKey',
+        'MinKey',
+        'ObjectId',
+        'ObjectID',
+        'BSONRegExp',
+        'Symbol',
+        'Timestamp'
+      ],
+      '../../../node_modules/bson/dist/bson.js': [
```
jsflax: [q] is this reliable?

edaniels: yes. That's the hoisted path, the one preceding it is the local path. As long as we are using browser, this will be there and the build will break otherwise.

jsflax: 👍

PR #146, opened 2018-06-22
STITCH-1474 Verify documentation and README are complete

adamchel:
This is the updated README, updated documentation, and docs generation script. To generate the docs, run `npm run docs`.


tdconfig.json
```diff
@@ -0,0 +1,26 @@
+{
```
edaniels: {
    "extends": "./config/tsconfig.esm.browser.base.json",
    "compilerOptions": {
    	"outDir": "dist/doc"
    },
    "include": [
      same include
    ],
    "exclude": [
        same exclude - lib
    ]
}

PR #154, opened 2018-06-28
STITCH-1720 - Implement Server (Node.js) SDK

edaniels:


packages/server/core/src/core/auth/internal/StitchAuthImpl.ts
```diff
@@ -0,0 +1,184 @@
...
+
+import version from "../../internal/common/Version";
+import AuthProviderClientFactory from "../providers/internal/AuthProviderClientFactory";
+import NamedAuthProviderClientFactory from "../providers/internal/NamedAuthProviderClientFactory";
+import StitchAuth from "../StitchAuth";
+import StitchAuthListener from "../StitchAuthListener";
+import StitchUser from "../StitchUser";
+import StitchUserFactoryImpl from "./StitchUserFactoryImpl";
+
+/**
```
adamchel: [nit] Don't think we need any of these `Partial` things.

packages/server/core/src/core/auth/internal/StitchAuthImpl.ts
```diff
@@ -0,0 +1,184 @@
...
+ * Partial JSDOM window.location containing only the
+ * properties and method we need
+ */
+interface PartialLocaion {
```
adamchel: [nit] assuming this same typo exists in browser, you should probably drive-by fix it there as well as delete here.

packages/server/core/src/core/auth/internal/StitchAuthImpl.ts
```diff
@@ -0,0 +1,184 @@
...
+      info[DeviceFields.DEVICE_ID] = this.deviceId;
+    }
+    if (this.appInfo.localAppName !== undefined) {
+      info[DeviceFields.APP_ID] = this.appInfo.localAppName;
+    }
+    if (this.appInfo.localAppVersion !== undefined) {
+      info[DeviceFields.APP_VERSION] = this.appInfo.localAppVersion;
+    }
+
+    info[DeviceFields.PLATFORM] = "server";
```
adamchel: [minor] shouldn't this be `node-server` or `js-server`?

packages/server/sdk/README.md
```diff
@@ -0,0 +1,172 @@
...
+    client.close();
+})
+```
+
+6. Run the app by running `node src/index.js`.
+
+### Using the SDK
+
+#### Initialize the SDK
+1. When your app or webpage is initialized, run the following code to initialize the Stitch SDK, replacing `<your-client-app-id>` with your Stitch application's client app ID:
```
adamchel: [nit] app or webpage -> Node.js service

packages/server/core/src/core/auth/internal/StitchAuthImpl.ts
edaniels: yes

PR #160, opened 2018-07-27
STITCH-1732 Create React Native SDK

adamchel:
This is ready to review, and almost ready to publish, I just need to update the READMEs.

Main things to note:

- The most relevant files to review in `mongodb-stitch-react-native-core` are `RNAsyncStorage.ts` and `Stitch.ts` (since we now have async client initialization), 
- React Native doesn't play nicely with hoisted lerna dependencies since its packager doesn't follow symlinks, and its Jest configuration assumes that `react-native` and several other packages are in a particular project's `node_modules` so I've added some packages to `nohoist` in `lerna.json`. This only affects the react native projects.
- None of our user-facing package will list `react-native` as a dependency, it is only a peer dependency of `mongodb-stitch-react-native-core`. React Native's packager doesn't work when the RN package exists in multiple places in a `node_modules` directory. For the purposes of tests, some packages list `react-native` as a devDependency.
- Jest now requires us to explicitly specify `"testURL": "http://localhost/"` if we want to use `localStorage` in tests.
- Some of the other funky stuff we're doing in the RN tests are specified here: https://github.com/kulshekhar/ts-jest#react-native
- We couldn't include an example project inside the lerna monorepo because of https://github.com/facebook/metro/issues/1. The example that will eventually be in the README for the react native SDK will serve as the example until docs comes up with something.


packages/core/sdk/__tests__/auth/StitchUserProfileUnitTests.ts
```diff
@@ -57,8 +57,8 @@ describe("StitchUserProfile", () => {
     expect(stitchUserProfileImpl.gender).toEqual(GENDER);
     expect(stitchUserProfileImpl.birthday).toEqual(BIRTHDAY);
     expect(stitchUserProfileImpl.pictureUrl).toEqual(PICTURE_URL);
-    expect(stitchUserProfileImpl.minAge).toEqual(Number.parseInt(MIN_AGE));
-    expect(stitchUserProfileImpl.maxAge).toEqual(Number.parseInt(MAX_AGE));
+    expect(stitchUserProfileImpl.minAge).toEqual(Number.parseInt(MIN_AGE, 10));
```
edaniels: Why did this change?

packages/react-native/core/src/core/Stitch.ts
```diff
@@ -0,0 +1,158 @@
...
+   */
+  public static hasAppClient(clientAppId: string): boolean {
+    return appClients[clientAppId] !== undefined;
+  }
+
+  /**
+   * Initializes the default StitchAppClient associated with the application.
+   * 
+   * @param clientAppId The desired clientAppId for the client.
+   * @param config Additional configuration options (optional).
```
edaniels: doc return type

packages/react-native/core/src/core/Stitch.ts
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  /**
+   * Initializes a new, non-default StitchAppClient associated with the 
+   * application.
+   * 
+   * @param clientAppId The desired clientAppId for the client.
+   * @param config Additional configuration options (optional).
+   */
+  public static initializeAppClient(
```
edaniels: doc return type

packages/react-native/core/src/index.ts
```diff
@@ -0,0 +1,108 @@
...
+  StitchCredential,
+  StitchRequestError,
+  StitchRequestErrorCode,
+  StitchServiceError,
+  StitchServiceErrorCode,
+  StitchUserProfile,
+  StitchUserIdentity,
+  Storage,
+  MemoryStorage,
+  RNAsyncStorage,
```
edaniels: Do you actually want to expose this to users?

packages/react-native/coretest/package.json
```diff
@@ -0,0 +1,79 @@
+{
+  "name": "mongodb-stitch-react-native-coretest",
+  "version": "4.0.9",
+  "description": "web core test",
```
edaniels: update descs in these new files

packages/react-native/core/src/index.ts
adamchel: No, good catch. I was using it when debugging earlier but I can get rid of it here.

packages/react-native/core/README.md
```diff
@@ -0,0 +1,5 @@
+# mongodb-stitch-react-native-core
+
+This package contains the React Native Core.
+
+**For more information on using this on a server, see the [mongodb-stitch-react-native-sdk](https://www.npmjs.com/package/mongodb-stitch-react-native-sdk) package.**
```
edaniels: on react native

PR #169, opened 2018-10-15
[STITCH-2027] Add SDK methods for accessing static hosting admin api endpoints

malexandert:


src/admin.js
```diff
@@ -143,22 +147,45 @@ export class StitchAdminClient extends StitchClient {
...
                 get: ()=> api._get(valueUrl),
                 remove: ()=> api._delete(valueUrl),
-                update: (data) => api._put(valueUrl, data)
+                update: (data) => api._put(valueUrl, { body: JSON.stringify(data) })
               };
             }
           }),
 
+          hosting: () => ({
+            config: () => ({
```
malexandert: The ticket specifies these, but they're not implemented yet in the backend. Should I just remove them for now?

mpobrien: you can leave em in for now. it'll just return a 404 error until it's used against a version of the server that has them implemented, but that's OK.

test/admin/hosting.test.js
```diff
@@ -0,0 +1,114 @@
...
+
+  it('listing assets when there are none should return an empty list', async() => {
+    let assets = await hosting.assets().list();
+    expect(assets).toHaveLength(0);
+  });
+
+  it('creating an asset should work', async() => {
+    let { metadata, body } = createTestFile();
+    await hosting.assets().create(JSON.stringify(metadata), body);
+    let assets = await hosting.assets().list({ recursive: true });
```
mpobrien: can we also add a test that uses the path prefix arg as well? (that'll be the form we will end up using most in the UI)

PR #177, opened 2018-11-20
STITCH-2214 package modules as CommonJS instead of UMD

adamchel:
Adding all reviewers for visibility but this change is pretty straightforward. I would like to get this reviewed ASAP so we can publish `4.1.0`. After this is committed, I will publish `4.1.0` and delete the STITCH-2214 branch from the main repository.

**Main Changes**:
* replace all UMD config files and UMD targets with CommonJS config files and CommonJS targets
* expose a CommonJS module for every `package.json` `"main"` field.
* updated React Native peer dependency to `^0.57.0`
* updated README to reflect new React Native `Getting Started` guide
* updated README to reflect upcoming `4.1.0` release

I published a `4.0.15-beta-1.0` release to test these changes and verified that:
* the React Native SDK now works out of the box with the new Expo Getting Started process for React Native
* the current workflow for using the browser SDK with webpack still works normally
* the current workflow for using the server SDK with Node.js still works normally

**Drive-bys**:
* Updated `"npm"` in root `package.json` to `6.4.1` to resolve vulnerabilities in dev dependency tree.


adamchel:
Also
* fixed some code in `FetchStreamTransport.ts` to be null-safe now that typescript has type information for `fetch`

PR #178, opened 2018-11-27
Fix logic condition in getAppClient

jc:


edaniels:
We will take a look at this tomorrow but nice find!

adamchel:
Hi @jc thanks for bringing this bug to our attention and for opening a PR to fix it. I'm running your change against our continuous integration suite. Once it passes I'll merge your fix and it will be released in our upcoming 4.1.0 release!

PR #180, opened 2018-11-27
STITCH-2008 - JS location metadata lookup, unit tests to follow

dkaminsky:
This is the Javascript SDK implementation of the logic for multi-region hostname lookup. Upon the first call to the client API, rather than directly accessing the underlying service, the SDK will now first await the result of a lookup to /client/v2/app/:clientAppId/location and use the hostname specified in the result as the hostname for the user request.

Differs slightly from the other SDKs in that it relies on failed promises rather than explicit error catching logic in the case of failure to retrieve or parse the location metadata.


packages/core/sdk/src/internal/net/ApiAppMetadata.ts
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  public toJSON(): object {
+    return {
+      [Fields.DEPLOYMENT_MODEL]: this.deploymentModel,
+      [Fields.LOCATION]: this.location,
+      [Fields.HOSTNAME]: this.hostname
+    };
+  }
+}
```
adamchel: [nit] add a newline at the end of the file

packages/core/sdk/src/internal/net/StitchAppRequestClient.ts
```diff
@@ -0,0 +1,66 @@
...
+      .withPath(this.routes.appMetadataRoute)
+      .build();
+
+    return super.doRequestToURL(request, this.baseUrl)
+      .then(resp => {
+        this.appMetadata = AppMetadata.fromJSON(EJSON.parse(resp.body));
+        return this.appMetadata;
+      });
+  }
+}
```
adamchel: [nit] newline

packages/core/sdk/src/internal/net/StitchAppRequestClient.ts
```diff
@@ -0,0 +1,66 @@
...
+    private appMetadata? : AppMetadata;
+
+    public constructor(clientAppId: string, baseUrl: string, transport: Transport) {
+    super(baseUrl, transport);
+    this.clientAppId = clientAppId;
+    this.routes = new StitchAppRoutes(clientAppId);
+  }
+
+  public doRequest(stitchReq: StitchRequest): Promise<Response> {
+    return this.initAppMetadata().then(metadata => super.doRequestToURL(stitchReq, metadata.hostname));
```
adamchel: [style] In the interest of readability of the chain of events, I think this could be something like:

```js
return this.initAppMetadata()
    .then(metadata => super.doRequestToURL(stitchReq, metadata.hostname));
```

packages/core/sdk/src/internal/net/StitchAppRequestClient.ts
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  public doStreamRequest(stitchReq: StitchRequest, open: boolean = true, retryRequest?: () => Promise<EventStream>): Promise<EventStream> {
+    return this.initAppMetadata().then(metadata => super.doStreamRequestToURL(stitchReq, metadata.hostname, open, retryRequest));
```
adamchel: [style] ditto

dkaminsky: ditto

PR #181, opened 2018-12-03
[STITCH-2183] Make support/3.x branch green again

malexandert:
Things still failing on Evergreen as of [this build](https://evergreen.mongodb.com/task/stitch_js_sdk_3.0_linux_64_run_tests_patch_d309cecb4a5ad12b9aaa714903b5a8ea9826edc5_5c05796f2fbabe111110e046_18_12_03_18_43_59):
- `test/test.js`
- `test/auth.test.js`
- `test/storage.test.js`
- ~`test/client/oauth.test.js`~

### DIFF
--- a/test/storage.test.js
+++ b/test/storage.test.js
@@ -42,11 +42,21 @@ function _runReverseMigration(toVersion, storage) {
   }
 }
 
+// The type of the item in storage differs between the MockBrowser and Evergreen
+function storageValue(value, storageType, isMocked) {
+  if (!isMocked && (storageType === 'localStorage' || storageType === 'sessionStorage')) {
+    return `${value}`;
+  }
+  return value;
+}
+
 describe('storage', function() {
   const namespace = 'client-test-app1';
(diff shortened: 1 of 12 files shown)


test/storage.test.js
```diff
@@ -42,11 +42,21 @@ function _runReverseMigration(toVersion, storage) {
   }
 }
 
+// The type of the item in storage differs between the MockBrowser and Evergreen
+function storageValue(value, storageType, isMocked) {
```
mpobrien: this is a bit odd - can you clarify a bit why it's different between MockBrowser and Evergreen? The "mock" should behave like the real thing as much as possible so it smells like there might be an underlying issue somewhere that we should address, if we can

malexandert:
Will follow up on auth-specific/mocking failures in a future ticket. [Here's](https://evergreen.mongodb.com/version/5c11598d2a60ed0573dc2514) a build with the remaining errors

PR #193, opened 2019-01-16
STITCH-2341 Export BSON library with SDKs at top level

adamchel:
Without the `package-lock.json` changes this PR is tiny.

Changes:
* imports `BSON from "bson"` and exports it at the `mongodb-stitch-core-sdk` level
* exports this `BSON` at the `mongodb-stitch-browser-core`, `mongodb-stitch-react-native-core`, and `mongodb-stitch-server-core` levels.
    * all of the `*-sdk` modules get this `BSON` automatically because those modules `import * from` their respective core module.
* changed `BSON` imports in integration tests to be from Stitch modules rather than from `"bson"`

I've also verified that `BSON` is usable in the generated `stitch.js` for use in script tags.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,5 +1,4 @@
 {
-	"name": "stitch-js-sdk",
 	"requires": true,
 	"lockfileVersion": 1,
 	"dependencies": {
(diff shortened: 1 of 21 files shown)


dkaminsky:
Lgtm

PR #206, opened 2019-01-28
STITCH-2206 Add watch to the RemoteMongoDBService at the collection level

adamchel:
This PR polishes support for `watch` on RemoteMongoCollection, and supports it in the `browser` and `server` SDKs. The server SDK uses a mature `eventsource` polyfill that is compatible with Node.js, and the browser SDK assumes that `EventSource` exists on the `window` object either natively or via a user-provided polyfill.

In addition to exposing and documenting `watch` on `RemoteMongoCollection`, this PR also exposes and documents the `Stream` and `StreamListener` types from the core SDK, since `watch` uses these types in its public-facing API.

As discussed offline and on Slack internally, we are not providing support for React Native due to the lack of a mature EventSource polyfill for React Native.


adamchel:
@jsflax PTAL

PR #214, opened 2019-02-01
DOCSP-4285 Add example to StitchAuthListener

cbush:
Adds a code example and some more rationale about the design choice
of not passing the event info.

Cloned changed to server and react-native.

### DIFF
--- a/packages/browser/core/src/core/auth/StitchAuthListener.ts
+++ b/packages/browser/core/src/core/auth/StitchAuthListener.ts
@@ -17,10 +17,39 @@
 import StitchAuth from "./StitchAuth";
 
 /**
- * An interface to be inherited by classes that need to take action whenever a 
- * particular {@link StitchAppClient} performs an authentication event. An 
- * instance of a {@link StitchAuthListener} must be registered with a 
- * {@link StitchAuth} for this to work correctly.
+ * StitchAuthListener is an interface for taking action whenever 
+ * a particular [[StitchAppClient]]'s authentication state changes.
+ *
+ * Implement [[onAuthEvent]] to handle these events. You can register 
+ * your listener with [[StitchAuth]] using [[StitchAuth.addAuthListener]].
+ *
+ * StitchAuth calls registered listeners when:
+ * - a user logs in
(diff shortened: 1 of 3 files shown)


packages/browser/core/src/core/auth/StitchAuthListener.ts
```diff
@@ -17,10 +17,39 @@
...
+ * a particular [[StitchAppClient]]'s authentication state changes.
+ *
+ * Implement [[onAuthEvent]] to handle these events. You can register 
+ * your listener with [[StitchAuth]] using [[StitchAuth.addAuthListener]].
+ *
+ * StitchAuth calls registered listeners when:
+ * - a user logs in
+ * - a user logs out
+ * - a user is linked to another identity
+ * - a listener is registered
```
adamchel: [task] add `* - a user is switched `. This will be true very soon when my multi-user PR goes in

packages/browser/core/src/core/auth/StitchAuthListener.ts
```diff
@@ -30,6 +59,16 @@ export default interface StitchAuthListener {
...
    *
+   * The [[StitchAuth]] instance itself is passed to this callback. This can be used to read the current state of authentication.
+   *
+   * ### Note
+   * Specific event details are deliberately not provided here because the events could be stale by the time they are handled.
+   * 
+   * For example, a user could log in then log out before the first login event is handled.
+   * 
+   * The intention is that you would treat this callback as a trigger to refresh the relevant parts of your app based
+   * on the new, current auth state. In other words, it should not be necessary to deduce exactly what changed in this callback.
```
adamchel: not sure if I agree with the "In other words, it should not be necessary to deduce exactly what changed". That's ultimately up to the app developer. If they want to maintain previous state so that they can compare to find out exactly what happened, that's totally fine. It might even help optimize whatever state they are refreshing.

cbush: Yeah, I think I went overly opinionated with that line, when what I really wanted to convey is already covered by the first line.

PR #216, opened 2019-02-04
STITCH-2483 basic support for multi-user authentication

adamchel:
This should cover STITCH-2483 and STITCH-2484.

I excluded `lastAuthActivity` because this PR was getting big and I wanted to get something up. I can do `lastAuthActivity` in a separate PR after this one is merged, as it is quite independent from the rest of this work.

This should cover everything in the new spec doc Tyler wrote, as well as preserving device ID when not logged in, which is something I'll add to the spec.

### DIFF
--- a/packages/browser/core/src/core/auth/internal/StitchAuthImpl.ts
+++ b/packages/browser/core/src/core/auth/internal/StitchAuthImpl.ts
@@ -125,8 +125,6 @@ export default class StitchAuthImpl extends CoreStitchAuth<StitchUser>
     }
   }
 
-
-
   public loginWithCredential(
     credential: StitchCredential
   ): Promise<StitchUser> {
--- a/packages/browser/core/src/core/auth/internal/StitchAuthImpl.ts
+++ b/packages/browser/core/src/core/auth/internal/StitchAuthImpl.ts
@@ -125,6 +125,8 @@ export default class StitchAuthImpl extends CoreStitchAuth<StitchUser>
     }
   }
 
+
+
   public loginWithCredential(
     credential: StitchCredential
   ): Promise<StitchUser> {
@@ -226,7 +228,19 @@ export default class StitchAuthImpl extends CoreStitchAuth<StitchUser>
   }
 
   public logout(): Promise<void> {
-    return Promise.resolve(super.logoutInternal());
+    return super.logoutInternal();
+  }
+
+  public logoutUserWithId(userId: string): Promise<void> {
+    return super.logoutUserWithIdInternal(userId);
+  }
+
+  public removeUser(): Promise<void> {
+    return super.removeUserInternal();
+  }
+
+  public removeUserWithId(userId: string): Promise<void> {
+    return super.removeUserWithIdInternal(userId);
   }
 
   protected get deviceInfo() {
--- a/packages/browser/coretest/__tests__/StitchAppClientIntTests.ts
+++ b/packages/browser/coretest/__tests__/StitchAppClientIntTests.ts
@@ -15,7 +15,7 @@
  */
 
 import { sign } from "jsonwebtoken";
-import { UserPasswordAuthProviderClient } from "mongodb-stitch-browser-core";
+import { UserPasswordAuthProviderClient, Stitch } from "mongodb-stitch-browser-core";
 import { BaseStitchBrowserIntTestHarness } from "mongodb-stitch-browser-testutils";
 import {
   Anon,
@@ -32,7 +32,8 @@ import {
   CustomCredential,
   UserPasswordAuthProvider,
   UserPasswordCredential,
-  UserType
+  UserType,
+  MemoryStorage
 } from "mongodb-stitch-core-sdk";
 
 const harness = new BaseStitchBrowserIntTestHarness();
@@ -81,7 +82,7 @@ describe("StitchAppClient", () => {
     expect(client.auth.isLoggedIn).toBeTruthy();
   });
 
-  it("should follow multiple login semantics", async () => {
+  it("should follow multiple login semantics and allow multiple users", async () => {
     const [appResponse, app] = await harness.createApp();
     await harness.addProvider(app as App, new Anon());
     await harness.addProvider(
@@ -93,7 +94,11 @@ describe("StitchAppClient", () => {
         "password subject"
       )
     );
-    const client = harness.getAppClient(appResponse as AppResponse);
+
+    const concreteAppResponse = appResponse as AppResponse;
+
+    let storage = new MemoryStorage(concreteAppResponse.clientAppId);
+    let client = harness.getAppClient(concreteAppResponse, storage);
 
     // check storage
     expect(client.auth.isLoggedIn).toBeFalsy();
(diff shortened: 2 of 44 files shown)


packages/browser/core/src/core/auth/internal/StitchAuthImpl.ts
```diff
@@ -125,6 +125,8 @@ export default class StitchAuthImpl extends CoreStitchAuth<StitchUser>
     }
   }
 
+
```
jsflax: [style] extra newlines

packages/core/sdk/src/auth/internal/CoreStitchAuth.ts
```diff
@@ -108,18 +121,28 @@ export default abstract class CoreStitchAuth<TStitchUser extends CoreStitchUser>
...
         StitchClientErrorCode.CouldNotLoadPersistedAuthInfo
       );
     }
-    if (info === undefined) {
-      this.authInfo = AuthInfo.empty();
+    if (activeUserAuthInfo === undefined) {
+      this.activeUserAuthInfo = AuthInfo.empty();
     } else {
-      this.authInfo = info;
+      this.activeUserAuthInfo = activeUserAuthInfo;
```
jsflax: [q][minor] Didn't we agree that we wanted to preserve the order of the list? In which case, a map would no longer be the appropriate data structure?

adamchel: I'm just using a plain JS object, which I believe is ordered. the order did seem to be preserved in my tests.

In any case, after my next PR, the user-facing list will be sorted by `lastAuthActivity` and `userId`, so the on-device order will not necessarily matter. I'll probably do a lazy sorting approach so the list won't need to be re-sorted every time it is requested

jsflax:
But I thought we came to the agreement that we are not sorting by either.
The conversation went that it would be a bad UX to not maintain the order
on the device– there is no reason for us to do the sorting for them that
they can do themselves.

Am Di., 5. Feb. 2019 um 15:55 Uhr schrieb Adam Chelminski <
notifications@github.com>:

adamchel:
@jsflax @dkaminsky @tkaye407 PTAL

In moving from plain JS objects to a map, I also had to enable the `downlevelIteration` compiler flag, since we are targeting ES5, and iterable Maps are an ES6 feature.

packages/core/sdk/__tests__/auth/internal/CoreStitchAuthUnitTests.ts
```diff
@@ -248,46 +253,285 @@ describe("CoreStitchAuthUnitTests", () => {
...
+    expect(auth.listUsers().length).toEqual(2);
+
+    auth.listUsers().forEach(user => {
+      expect(user.isLoggedIn).toBeFalsy();
+    });
+
+    expect(auth.isLoggedIn).toBeFalsy();
+
+    // assert that we still have a device ID 
+    expect(auth.hasDeviceId).toBeTruthy();
```
tkaye407: @jsflax this is the test case

jsflax:
LGTM!

packages/browser/coretest/__tests__/StitchAppClientIntTests.ts
```diff
@@ -154,6 +159,98 @@ describe("StitchAppClient", () => {
...
+    expect(client.auth.listUsers().length).toEqual(3);
+    
+    // verify ordering is preserved
+    expect(client.auth.listUsers()[0].id).toEqual(anonUser.id);
+    expect(client.auth.listUsers()[1].id).toEqual(emailUserId);
+    expect(client.auth.listUsers()[2].id).toEqual(id2);
+
+    // verify that removing the user with id2 removes the second email/pass user
+    // and logs out the active user
+    await client.auth.logoutUserWithId(id2);
```
tkaye407: For caution, id add a test whenever you log out to check that the array persisted it: expect(client.auth.listUsers()[2].isLoggedIn).toEqual(false)

packages/core/sdk/__tests__/auth/internal/CoreStitchAuthUnitTests.ts
```diff
@@ -248,46 +253,285 @@ describe("CoreStitchAuthUnitTests", () => {
...
 
-        const [actualRequest] = capture(
-          requestClientMock.doRequest
-        ).byCallIndex(2);
-        expect(expectedRequest.build()).toEqualRequest(actualRequest);
+    await auth.logoutInternal();
 
-        expect(auth.isLoggedIn).toBeFalsy();
-      });
+    expect(auth.isLoggedIn).toBeFalsy();
```
tkaye407: I would try testing how some of these methods fail. Like how calling switch/logout with a non-existent user id should throw a userNotFound

packages/core/sdk/__tests__/auth/internal/CoreStitchAuthUnitTests.ts
```diff
@@ -248,46 +253,285 @@ describe("CoreStitchAuthUnitTests", () => {
...
+    const requestClient = instance(requestClientMock);
+    const routes = new StitchAppRoutes("my_app-12345").authRoutes;
+    const storage = new MemoryStorage(appId);
+    const auth = new StitchAuth(
+      requestClient,
+      routes,
+      storage
+    );
+
+    try {
```
tkaye407: re my comment above, I see that you handle failures here, but may still be nice to do it for all of the methods that throw

packages/core/sdk/__tests__/auth/internal/CoreStitchAuthUnitTests.ts
```diff
@@ -248,46 +253,285 @@ describe("CoreStitchAuthUnitTests", () => {
...
+
+    const expectedRequest = new StitchRequest.Builder();
+    expectedRequest.withMethod(Method.DELETE).withPath(routes.sessionRoute);
+    const headers = {
+      [Headers.AUTHORIZATION]: Headers.getAuthorizationBearer(
+        TEST_REFRESH_TOKEN
+      )
+    };
+    expectedRequest.withHeaders(headers);
+
```
tkaye407: nice, was just about to suggest this!

packages/core/sdk/src/auth/internal/CoreStitchAuth.ts
```diff
@@ -701,4 +898,39 @@ export default abstract class CoreStitchAuth<TStitchUser extends CoreStitchUser>
...
+    if (unclearedAuthInfo === undefined) {
+      // this doesn't necessarily mean there's an error. we could be in a 
+      // provisional state where the profile request failed and we're just
+      // trying to log out the active user.
+      if (this.activeUserAuthInfo.userId !== userId) {
+        // only throw if this ID is not the active user either
+        throw new StitchClientError(StitchClientErrorCode.UserNotFound);  
+      }
+    }
+
```
tkaye407: This kind of seems like weird behaviour / should change the name of this function. When I hear clearUserAuth I assume that this method would delete the user entirely from the list of users, not just log them out

packages/core/sdk/src/auth/internal/models/StoreAuthInfo.ts
```diff
@@ -58,11 +92,46 @@ function writeToStorage(authInfo: AuthInfo, storage: Storage) {
...
+  storage.set(
+    StoreAuthInfo.ALL_USERS_STORAGE_NAME, 
+    JSON.stringify(encodedStoreInfos)
+  );
 }
 
 class StoreAuthInfo extends AuthInfo {
-  public static readonly STORAGE_NAME: string = "auth_info";
+  public static readonly ACTIVE_USER_STORAGE_NAME: string = "auth_info";
+  public static readonly ALL_USERS_STORAGE_NAME
```
tkaye407: Where is this set?

adamchel: nowhere, good catch

packages/core/sdk/__tests__/auth/internal/CoreStitchAuthUnitTests.ts
adamchel: I test the switch case in the switch test but I'll add the logout case

packages/core/sdk/src/auth/internal/CoreStitchAuth.ts
adamchel: per the other PR, I think `clearUserAuthTokens` is the most appropriate

PR #219, opened 2019-02-07
DOCSP-4283 Validate code examples

cbush:
- Externalizes code examples in contrib/docs-examples/snippets.
- Adds validator script. Not so much a unit test as just running
eslint and the script itself to check that the script is formatted
and the APIs are still valid, etc.
- Adds two TypeDoc plugins: code-examples and tags-by-name
- code-examples plugin pastes code examples from external files to
the comment tags of a matching-named entity
- tags-by-name reorders the TypeDoc tags of each entity to be more
usable by the custom TypeDoc theme.
- Updates the custom theme to use the new examples tags and also
the "See also" tag.
- Updates all code examples to use @example tag or remove (externalize) the example.
- Updates comments to use @see instead of See Also.

Staged changes:

https://docs-mongodbcom-staging.corp.mongodb.com/stitch/bush/docsp-4283-code-example-validation/sdk/js/index.html

https://docs-mongodbcom-staging.corp.mongodb.com/stitch/bush/docsp-4283-code-example-validation/sdk/js-server/index.html

https://docs-mongodbcom-staging.corp.mongodb.com/stitch/bush/docsp-4283-code-example-validation/sdk/react-native/index.html


package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "stitch-js-sdk",
+	"name": "stitch-js-sdk",
```
adamchel: [nit] I can't tell if these are tabs or spaces from the Github interface, but make sure it's consistent across the whole file.

package.json
```diff
@@ -24,7 +24,7 @@
 	},
 	"scripts": {
 		"docs": "npm run docs-theme && npm run docs-browser && npm run docs-server && npm run docs-react-native",
-    "docs-theme": "pushd typedoc-theme && npm run build",
+		"docs-theme": "pushd typedoc-theme && npm run build",
```
adamchel: [nit] ditto

cbush: You're right. Good catch! But I think this change actually makes it consistent now.

contrib/docs-examples/validate.js
```diff
@@ -0,0 +1,35 @@
+
+// ^^^ Paste example() definition ^^^
```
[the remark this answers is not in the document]
cbush: Examples are pasted automatically by the validate.sh script. Will add note!

PR #230, opened 2019-02-14
STITCH 2497: Remove core dependency on cross-fetch

tkaye407:
Added in a few quick fixes for alphabetizing the export lists. Furthermore, not sure if we should test to make sure it works properly in react native


packages/browser/core/src/core/internal/net/BrowserFetchStreamTransport.ts
```diff
@@ -16,40 +16,19 @@
...
-      const response = values[0];
-      const body = values[1];
-      const headers: { [key: string]: string } = {};
-      response.headers.forEach((value, key) => {
-        headers[key] = value;
-      });
-      return new Response(headers, response.status, body);
-    });
-  }
+export default class BrowserFetchStreamTransport extends BrowserFetchTransport {
```
adamchel: [nice] good call

packages/browser/core/src/core/internal/net/BrowserFetchTransport.ts
```diff
@@ -0,0 +1,56 @@
...
+  BasicRequest,
+  ContentTypes,
+  EventStream,
+  Headers,
+  Response,
+  Transport,
+  StitchClientError, 
+  StitchClientErrorCode
+} from "mongodb-stitch-core-sdk";
+import {fetch as fetch} from 'whatwg-fetch'
```
adamchel: [nit] add a line between the last import and the doc comment

packages/core/sdk/__tests__/internal/net/StitchRequestClientUnitTests.ts
```diff
@@ -26,12 +26,26 @@ import { StitchRequest } from "../../../src/internal/net/StitchRequest";
 import StitchError from "../../../src/StitchError";
 import { StitchServiceErrorCode } from "../../../src/StitchServiceErrorCode";
 import StitchServiceError from "../../../src/StitchServiceError";
+import { promises } from "fs";
+/** @hidden */
+// TODO: mockito does not yet support mocking interfaces but it will soon
```
adamchel: [nit] I don't think this needs to be a TODO in the code. Get rid of the `TODO: ` but keep the rest of the comment. You can file a ticket, but I don't think it's necessary because this is just test code.

packages/core/testutils/src/JestFetchStreamTransport.ts
```diff
@@ -0,0 +1,72 @@
...
+          error => reject(error),
+          retryRequest ? 
+            () => retryRequest().then(es => es as JestEventSourceEventStream)
+            : undefined
+          )
+      );
+      /* eslint-enable no-unused-expressions */
+    });
+  }
+}
```
adamchel: [nit] add a newline at the end of the file

packages/core/testutils/src/JestEventSourceEventStream.ts
```diff
@@ -0,0 +1,87 @@
...
+import {
+  BaseEventStream,
+  Event,
+  StitchClientError,
+  StitchClientErrorCode, 
+  StitchEvent
+} from "mongodb-stitch-core-sdk";
+
+/** @hidden */
+export default class JestEventSourceEventStream extends BaseEventStream<JestEventSourceEventStream> {
```
adamchel: [broader discussion] So this works and I know it's what I told you to do, but now this means that the Transports we wrote are not covered at all by our tests unless we do E2E browser, Node.js, and React Native testing.

We aren't going to make changes to the other transports again for a while probably, but it might be a good idea to file a backlog ticket to do some very rudimentary E2E smoke testing with an actual node server, actual react native app, and actual browser (via something like selenium) to make sure the transports properly interact with the network.

adamchel: thoughts @jsflax ?

packages/core/testutils/src/JestFetchTransport.ts
```diff
@@ -0,0 +1,57 @@
...
+        headers[key] = value;
+      });
+      return new Response(headers, response.status, body);
+    });
+  }
+
+  public stream(request: BasicRequest, open = true, retryRequest?: () => Promise<EventStream>): Promise<EventStream> {
+    throw new StitchClientError(StitchClientErrorCode.StreamingNotSupported);
+  }
+}
```
adamchel: [nit] newline

packages/react-native/core/src/core/internal/net/ReactNativeFetchTransport.ts
```diff
@@ -0,0 +1,57 @@
...
+        });
+        return new Response(headers, response.status, body);
+      });
+    }
+  
+    public stream(request: BasicRequest, open: boolean = true, retryRequest?: () => Promise<EventStream>): Promise<EventStream> {
+      throw new StitchClientError(StitchClientErrorCode.StreamingNotSupported);
+    }
+  }
+
```
adamchel: [nit] get rid of the whitespace so the last line of the file is just a newline

packages/react-native/core/src/core/internal/net/ReactNativeFetchTransport.ts
```diff
@@ -0,0 +1,57 @@
...
+    EventStream,
+    Headers,
+    Response,
+    Transport,
+    StitchClientError, 
+    StitchClientErrorCode
+  } from "mongodb-stitch-core-sdk";
+  
+  /** @hidden */
+  export default class ReactNativeFetchTransport implements Transport {
```
adamchel: [nit] rename this class to `RNFetchTransport`

packages/core/testutils/src/JestEventSourceEventStream.ts
jsflax: Consider how core of an element the transports are, I think we should add some basic tests to make sure everything still works, before merging this. More can be added later.

adamchel: The point is that these transports can't be tested in Jest, so we'd have to create an entire E2E testing strategy, which I feel is out of scope for this ticket, which was meant to fix a bug.

Tyler did do some manual E2E in the browser to make sure the transports still work, and I think it would be good to merge if he did the same for Server and RN

jsflax: Ah, fair point. That works for me.

PR #233, opened 2019-02-16
Add findOne() functionality

tkaye407:
Not sure if this is a big deal/wrong, but the findOne() method returns null as opposed to undefined when there are no matching documents. We can discuss if this is not the right behaviour.

### DIFF
--- a/packages/browser/services/mongodb-remote/src/RemoteMongoCollection.ts
+++ b/packages/browser/services/mongodb-remote/src/RemoteMongoCollection.ts
@@ -104,12 +104,12 @@ export default interface RemoteMongoCollection<DocumentT> {
    * An empty query (`{}`) will match all documents.
    *
    * @param query the query filter
-   * @return the resulting document or undefined if the query resulted in zero matches
+   * @return the resulting document or null if the query resulted in zero matches
    */
   findOne(
     query?: object,
     options?: RemoteFindOptions
-  ): Promise<DocumentT | undefined>;
+  ): Promise<DocumentT | null>;
 
   /**
    * Aggregates documents according to the specified aggregation pipeline.
--- a/packages/browser/services/mongodb-remote/src/RemoteMongoCollection.ts
+++ b/packages/browser/services/mongodb-remote/src/RemoteMongoCollection.ts
@@ -99,6 +99,19 @@ export default interface RemoteMongoCollection<DocumentT> {
   ): RemoteMongoReadOperation<DocumentT>;
 
   /**
+   * Finds one document in the collection that matches the given query.
+   * 
+   * An empty query (`{}`) will match all documents.
+   *
+   * @param query the query filter
+   * @return the resulting document or undefined if the query resulted in zero matches
+   */
+  findOne(
+    query?: object,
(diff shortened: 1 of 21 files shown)


packages/browser/services/mongodb-remote/src/RemoteMongoCollection.ts
```diff
@@ -98,6 +98,19 @@ export default interface RemoteMongoCollection<DocumentT> {
...
+   * 
+   * An empty query (`{}`) will match all documents.
+   *
+   * @param query the query filter
+   * @return the resulting document or undefined if the query resulted in zero matches
+   */
+  findOne(
+    query?: object,
+    options?: RemoteFindOptions
+  ): Promise<DocumentT | undefined>;
```
adamchel: [minor] if `findOne` is returning null, the type here should be `Promise<DocumentT | null>`, but we should have a discussion in Slack about that

packages/browser/services/mongodb-remote/src/internal/RemoteMongoCollectionImpl.ts
```diff
@@ -82,6 +82,19 @@ export default class RemoteMongoCollectionImpl<DocumentT> {
...
+  /**
+   * Finds one document in the collection.
+   *
+   * @param query the query filter
+   * @return the resulting document or undefined if the query resulted in zero matches
+   */
+  public findOne(
+    query?: object,
+    options?: RemoteFindOptions
+  ): Promise<DocumentT | undefined> {
```
adamchel: [minor] ditto re: null

packages/core/services/mongodb-remote/src/internal/CoreRemoteMongoCollection.ts
```diff
@@ -65,6 +65,17 @@ export default interface CoreRemoteMongoCollection<DocumentT> {
...
+  /**
+   * Finds one documents in the collection.
+   *
+   * @param query the query filter
+   * @return the resulting document or undefined if no such document exists
+   */
+  findOne(
+    query?: object,
+    options?: RemoteFindOptions
+  ): Promise<DocumentT | undefined>;
```
adamchel: [nit] ditto re: null

packages/core/sdk/src/auth/internal/CoreStitchAuth.ts
```diff
@@ -211,6 +211,10 @@ export default abstract class CoreStitchAuth<TStitchUser extends CoreStitchUser>
   ): Promise<T> {
     return this.doAuthenticatedRequest(stitchReq)
       .then(response => {
+        if (response.body! === '{"$undefined":true}') {
+          return undefined;
+        }
+       
         const obj = EJSON.parse(response.body!, { strict: false });
```
tkaye407: This is what I added, let me know if this is too sketchy

adamchel: This is very sketchy, and I don't think it's a check we should be making for every request. 

Would it be better to do the following

```
if (obj === null) {
    return undefined;
}
```

after the EJSON.parse call?

PR #234, opened 2019-02-20
fix HTTP service test to accomodate httpbin.org bug

adamchel:


packages/browser/services/http/__tests__/HttpServiceClientIntTests.ts
```diff
@@ -69,7 +69,11 @@ describe("HttpServiceClient", () => {
     cookies.bob = "barker";
     const form = {};
     const headers = {};
-    headers.myHeader = ["value1", "value2"];
+    headers.Myheader = ["value1", "value2"];
+
+    // due to httpbin bug that doesn't return Content-Length without these
```
tkaye407: It seems you failed evergreen lint because these comments (in all 3 files) have to be capitalized

PR #235, opened 2019-02-20
STITCH-2539: Fix the JS-Linter

tkaye407:


packages/core/sdk/src/internal/common/mongodb-extjson.d.ts
```diff
@@ -15,18 +15,18 @@
  */
```
tkaye407: Some interesting changes on this file, it seems auto-fix did the same thing as above here with the type / interface

tkaye407: Looking online, it seems like it it better to use interfaces as opposed to types which seem to just be aliases

packages/core/services/http/src/HttpRequest.ts
```diff
@@ -167,11 +167,11 @@ export namespace HttpRequest {
      * @return the built HTTP request.
      */
     public build(): HttpRequest {
-      if (this.url == null || this.url === "") {
+      if (this.url === undefined || this.url === "") {
         throw new Error("must set url");
```
tkaye407: I found this to be interesting, lint didnt like comparing to null

packages/react-native/core/src/core/auth/internal/StitchAuthImpl.ts
```diff
@@ -170,10 +170,9 @@ export default class StitchAuthImpl extends CoreStitchAuth<StitchUser>
    */
   public onAuthEvent(listener?: StitchAuthListener) {
     if (listener) {
-      const auth = this;
       const _ = new Promise(resolve => {
         if (listener.onAuthEvent) {
-          listener.onAuthEvent(auth);  
+          listener.onAuthEvent(this);  
         }
```
tkaye407: Just making sure passing the object directly has no impact, shouldnt be any different than assigning it to the variable auth though

packages/server/core/src/core/auth/internal/StitchAuthImpl.ts
```diff
@@ -170,10 +170,9 @@ export default class StitchAuthImpl extends CoreStitchAuth<StitchUser>
    */
   public onAuthEvent(listener?: StitchAuthListener) {
     if (listener) {
-      const auth = this;
       const _ = new Promise(resolve => {
         if (listener.onAuthEvent) {
-          listener.onAuthEvent(auth);  
+          listener.onAuthEvent(this);  
         }
```
tkaye407: Same as above

packages/browser/core/src/core/internal/net/BrowserFetchStreamTransport.ts
```diff
@@ -16,52 +16,55 @@
...
-        headers[key] = value;
+        respHeaders[key] = value;
       });
       if (response.status < 200 || response.status >= 300) {
         return response.text()
-        .then(body => handleRequestError(new Response(headers, response.status, body)));
+        .then(body => handleRequestError(new Response(respHeaders, response.status, body)));
       }
 
-      return new Promise<EventStream>((resolve, reject) => {
```
tkaye407: Couldnt find a way around this without breaking things

packages/browser/core/src/core/auth/internal/StitchAuthImpl.ts
```diff
@@ -308,10 +306,9 @@ export default class StitchAuthImpl extends CoreStitchAuth<StitchUser>
    */
   public onAuthEvent(listener?: StitchAuthListener) {
     if (listener) {
-      const auth = this;
       const _ = new Promise(resolve => {
         if (listener.onAuthEvent) {
-          listener.onAuthEvent(auth);  
+          listener.onAuthEvent(this);  
         }
```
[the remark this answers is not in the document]
adamchel: I think the original pattern came from Swift where memory management is a bit different, so I think this is fine

packages/browser/core/src/core/auth/internal/StitchAuthImpl.ts
```diff
@@ -423,7 +421,7 @@ export default class StitchAuthImpl extends CoreStitchAuth<StitchUser>
   }
 
   private cleanupRedirect() {
-    this.jsdomWindow.history.replaceState(null, "", this.pageRootUrl());
+    this.jsdomWindow.history.replaceState(undefined, "", this.pageRootUrl());
```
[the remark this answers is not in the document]
adamchel: Hmm this one is tricky. I can't tell from Googling if this will work. It will _probably_ work, but to be sure you should set up an E2E scenario with Google redirect auth, and see if the auth tokens are cleared from the URL after redirect auth is complete. If you need help with this lmk, I have a pretty good idea of what you need to do.

packages/browser/core/src/core/auth/internal/StitchAuthImpl.ts
```diff
@@ -433,7 +431,7 @@ export default class StitchAuthImpl extends CoreStitchAuth<StitchUser>
   private parseRedirect(): ParsedRedirectFragment | never {
     if (typeof this.jsdomWindow === "undefined") {
       // This means we're running in some environment other
-      // than a browser - so handling a redirect makes no sense here.
+      // Than a browser - so handling a redirect makes no sense here.
```
adamchel: [aside] IMO this is a dumb rule because it makes the comments more confusing if they span multiple lines and they're not multiline comments.

I would be in favor of disabling this rule globally but let's have a discussion with the whole team.

packages/browser/core/src/core/internal/net/BrowserFetchStreamTransport.ts
adamchel: 👍 just add a comment explaining why you disabled the rule since this is not test code

packages/server/core/src/core/internal/net/NodeFetchStreamTransport.ts
```diff
@@ -52,42 +53,41 @@ export default class NodeFetchStreamTransport implements Transport {
...
-        .then(body => handleRequestError(new Response(headers, response.status, body)));
+        .then(body => handleRequestError(new Response(respHeaders, response.status, body)));
       }
 
-      return new Promise<EventStream>((resolve, reject) => {
-        // Any error against opening this stream here will have an error
-        // telling the user to look at the network response which is why we
-        // prefer the ReadableStream approach. Beyond that, this type of
-        // EventStream works the same and can still reconnect.
+      /* eslint-disable no-unused-expressions */
```
adamchel: [nit] add the comment here explaining why the rule is disabled

packages/core/services/http/src/HttpRequest.ts
```diff
@@ -51,7 +51,7 @@ export namespace HttpRequest {
     /**
      * Constructs a new builder for an HTTP request.
      */
-    public constructor() {}
+    public constructor() {/* */}
```
adamchel: [nit] what is this?

packages/core/services/aws/src/AwsRequest.ts
```diff
@@ -40,7 +40,7 @@ export namespace AwsRequest {
     /**
      * Constructs a new builder for an AWS request.
      */
-    public constructor() {}
+    public constructor() {/* */}
```
adamchel: [nit] ditto

packages/core/sdk/src/services/internal/CoreStitchServiceClientImpl.ts
```diff
@@ -70,9 +70,9 @@ export default class CoreStitchServiceClientImpl
   ): StitchAuthRequest {
     const body = { name };
     if (this.serviceName !== undefined) {
-      body["service"] = this.serviceName;
+      body["service".toString()] = this.serviceName;
```
adamchel: [nit] this seems weird, why is this necessary? could we perhaps improve this by making the type better?

packages/core/sdk/src/internal/net/BasicRequest.ts
```diff
@@ -85,9 +85,11 @@ export namespace BasicRequest {
       return new BasicRequest(
         this.method,
         this.url,
+        /* tslint:disable no-object-literal-type-assertion */
```
adamchel: [nit] add comment explaining why you disabled the rule/why we can't resolve this by making the types better

packages/core/sdk/src/internal/common/Base64.ts
```diff
@@ -16,8 +16,9 @@
 
 import { fromByteArray, toByteArray } from "base64-js";
 
-// sourced from https://github.com/coolaj86/TextEncoderLite
+// Sourced from https://github.com/coolaj86/TextEncoderLite
 /** @hidden */
+/* tslint:disable */
```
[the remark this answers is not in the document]
adamchel: sounds good to me, though we should probably include the license of where we sourced this from

packages/core/admin-client/src/services/rules/RulesResources.ts
```diff
@@ -87,7 +87,7 @@ export class RuleCreatorCodec implements Encoder<RuleCreator> {
 
 export class MongoDbCodec implements Encoder<MongoDbRuleCreator> {
   public encode(from: MongoDbRuleCreator): object {
-    from.rule["namespace"] = from.namespace;
+    from.rule["namespace".toString()] = from.namespace;
     return from.rule;
```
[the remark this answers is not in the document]
adamchel: what exactly is the rationale that the linter gives? perhaps we could set up the type better?

PR #238, opened 2019-02-26
STITCH-2546 Generate error if user ID is passed to destructive active user functions

adamchel:
Fear not the large number of additions and deletions, for they are only changes to `package-lock.json`. I drive-by patched some (but not all) local dev environment npm dependency vulns with `npm audit fix`.


tkaye407:
LGTM. I would debate passing a more descriptive error, but if you think UnexpectedArguments is enough then sounds good to me

PR #241, opened 2019-03-04
 Wrap switchUserWithId in try catch

bibyzan:
If switchUserWithId throws one of its potential errors within the Promise.resolve, it goes unhandled and crashes the page.


adamchel:
@bibyzan thanks for submitting this PR, good catch! It looks like it's passed our integration tests, so I'll go ahead and merge this now.

PR #248, opened 2019-03-08
STITCH-2617 - add optional filter paramater for apps.list function

chickenbug:
I added the filter parameter so that our adminSDK can filter by product when listing apps.


chickenbug:
This will need to go in after [STITCH-2611]( https://jira.mongodb.org/browse/STITCH-2611) which adds the atlas project to the test configs

PR #250, opened 2019-03-12
STITCH-2420 Add sdk support for deployment history api

marcgurney:
will add a commit with the dist files after it's approved


test/admin/deploy.test.js
```diff
@@ -0,0 +1,25 @@
...
+
+  beforeEach(async() => {
+    const { apiKey, groupId, serverUrl } = extractTestFixtureDataPoints(test);
+    th = await buildAdminTestHarness(true, apiKey, groupId, serverUrl);
+    appDeploy = th.app().deploy();
+  });
+
+  afterEach(async() => th.cleanup());
+
+  it('listing deploy history should return empty list', async() => {
```
UnicodeSnowman: make this read like a sentence: `it('returns an empty list when requesting deploy history', ...)`

test/admin/deploy.test.js
```diff
@@ -0,0 +1,25 @@
...
+    let deployHistory = await appDeploy.deployments().list();
```
UnicodeSnowman: `s/let/const`

PR #251, opened 2019-03-13
Stitch-2630: Requests without Auth should Reject Promise

tkaye407:
Are we worried about this being a breaking change for people's code?


packages/core/sdk/src/auth/internal/CoreStitchAuth.ts
```diff
@@ -559,9 +562,9 @@ export default abstract class CoreStitchAuth<TStitchUser extends CoreStitchUser>
   private prepareAuthRequest(
     stitchReq: StitchAuthRequest,
     authInfo: AuthInfo
-  ): StitchRequest {
+  ): Promise<StitchRequest> {
```
jsflax: [minor] It seems like you converted this to a `Promise` so that it plays nicer with your rejection logic above, but semantically it doesn't really make sense, since nothing async is actually happening here.

This should be remade synchronous and the logic above should be altered.

PR #253, opened 2019-03-15
DOCSP-4883 Document OAuth2 login with React Native

cbush:
This adds an example of how to use a 3rd-party module to
work with Google auth and Stitch.

Calls out differences between working with Browser SDK (RedirectCredential)
and server/RN.

Attempt to explain the general idea of how to work with OAuth2 auth
providers.

Staged:
https://docs-mongodbcom-staging.corp.mongodb.com/stitch/bush/docsp-4883-rn-google-login/sdk/js-server/index.html
https://docs-mongodbcom-staging.corp.mongodb.com/stitch/bush/docsp-4883-rn-google-login/sdk/react-native/index.html


contrib/docs-examples/snippets/react-native/GoogleCredential.js
```diff
@@ -0,0 +1,98 @@
...
+  // NOTE: The react-native-google-signin native module must be installed correctly
+  // and your Google project must be configured.
+  //
+  // For more detail, see https://stackoverflow.com/questions/55145071/#55173164
+  //
+  // Above:
+  // import {Stitch, GoogleCredential} from 'mongodb-stitch-react-native-sdk'
+  // import {GoogleSignin, GoogleSigninButton} from 'react-native-google-signin'
+  class GoogleLogin extends React.Component {
+    // ...
```
nlarew: nit: maybe we should be more explicit here?

e.g. `... other component methods`

contrib/docs-examples/snippets/react-native/GoogleCredential.js
```diff
@@ -0,0 +1,98 @@
...
+
+    componentDidMount() {
+      // Configure react-native-google-signin
+      GoogleSignin.configure({
+        webClientId: '<id>', // client ID of type WEB from Stitch
```
nlarew: should this be "from Stitch"? In the SO post you mention that it's the Client ID value from the Stitch UI. Since we don't have a screenshot handy here, maybe we should say that it's the client ID that you create as part of google project setup?

contrib/docs-examples/snippets/react-native/GoogleCredential.js
```diff
@@ -0,0 +1,98 @@
...
+        offlineAccess: true, // allows Stitch service to use credential
```
nlarew: I think I know what you mean here but I'm not sure. Is it something like: ```Must be `true` for the GoogleCredential to be valid```

contrib/docs-examples/snippets/react-native/GoogleCredential.js
```diff
@@ -0,0 +1,98 @@
...
+      // It's recommended to call this before signIn()
+      await GoogleSignin.hasPlayServices()
+
+      // Sign in via react-native-google-signin
+      const userInfo = await GoogleSignin.signIn()
+
+      // Retrieve the server auth code
+      const {serverAuthCode} = userInfo
+      if (serverAuthCode === null) {
+        throw new Error('Failed to get serverAuthCode!')
```
nlarew: Does `serverAuthCode === null` imply that the login failed or was cancelled? Could add a comment clarifying the case that would cause the error.

contrib/docs-examples/snippets/react-native/StitchAuth.loginWithCredential.js
```diff
@@ -0,0 +1,98 @@
+function example() {} function untested({View, Text, Button, React, GoogleSignin, Stitch, GoogleCredential, GoogleSigninButton}) { // too much boilerplate to be able to test this
```
nlarew: I get the reasoning for this change, but it seems incomplete to not include the anonymous and email/password login snippets here too. Maybe later down the road we can look into some kind of tabbed experience in the SDK docs to pick the credential you want an example for so that we can include them here as well as in StitchAuth?

packages/react-native/core/src/core/Stitch.ts
```diff
@@ -30,10 +30,14 @@ let appClients: { [key: string]: StitchAppClientImpl } = {};
  * Singleton class with static utility functions for initializing a [[StitchAppClient]].
  *
  * Typically, the [[Stitch.initializeDefaultAppClient]] method is all you need 
- * to instantiate the client:
+ * to instantiate the client (note that, unlike the other JS SDKs, this method returns
+ * a Promise):
```
nlarew: nit: maybe make it a bit more explicit which SDK isn't included in "the other JS SDKs"

"unlike in the other JS SDKs, this method returns  a Promise _in the React Native SDK_"?

contrib/docs-examples/snippets/react-native/StitchAuth.loginWithCredential.js
cbush: Don't worry, those snippets are still available elsewhere in the docs!

PR #254, opened 2019-03-25
DOCSP-4862 Add script to update browser sdk README versions

cbush:
Please run contrib/update_browser_readme_version.sh when releasing/publishing
docs.

### DIFF
--- a/contrib/README.md
+++ b/contrib/README.md
@@ -36,8 +36,11 @@ The [aws-cli](https://docs.aws.amazon.com/cli/latest/userguide/install-bundle.ht
 along with relevant permissions to the stitch-sdks bucket.
 
 ```bash
-lerna publish --force-publish="*"
 ./update_browser_readme_version.sh
+git add ../packages/browser/sdk/README.md
+git commit -m "Update SDK version in browser README"
+
+lerna publish --force-publish="*"
 ./publish_bundles.sh
 ./publish_docs.sh
 ```
--- a/contrib/README.md
+++ b/contrib/README.md
@@ -45,6 +45,9 @@ lerna publish --force-publish="*"
 ./publish_docs.sh
 ```
 
+Note that update_browser_readme_version.sh will git commit its changes to the README automatically.
+It will fail out if there are currently other changes to tracked files.
+
 The `--force-publish="*"` argument ensures that all packages bump their version in lockstep.
 
 ### Patch Versions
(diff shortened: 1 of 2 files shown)


contrib/README.md
```diff
@@ -37,6 +37,7 @@ along with relevant permissions to the stitch-sdks bucket.
 
 ```bash
 lerna publish --force-publish="*"
+./update_browser_readme_version.sh
```
adamchel: [minor] this should go before `lerna publish`, otherwise the READMEs on npmjs.com will show an old version.

contrib/update_browser_readme_version.sh
```diff
@@ -0,0 +1,18 @@
...
+VERSION_PATCH=$(echo $VERSION | cut -d. -f3 | cut -d- -f1)
+
+echo "Setting README version to $VERSION_MAJOR.$VERSION_MINOR.$VERSION_PATCH..."
+
+# Replace the occurences of stitch-sdks/js/bundles/<SOME VERSION>/stitch
+# with stitch-sdks/js/bundles/<CURRENT VERSION>/stitch
+sed -i '' \
+  -e "s#\(stitch-sdks/js/bundles/\)[0-9\.]*\(/stitch\)#\1$VERSION_MAJOR.$VERSION_MINOR.$VERSION_PATCH\2#g" \
+  ../packages/browser/sdk/README.md
+echo "README updated."
```
adamchel: [minor] To make sure this gets committed before a publish you should add 

```
git add ../packages/browser/sdk/README.md
git commit -m "Update SDK version to $VERSION_MAJOR.$VERSION_MINOR.$VERSION_PATCH in relevant READMEs"
```

cbush: I'm not sure how I feel about adding git commands in a script... Wouldn't this violate the principle of least astonishment? I wouldn't expect a script to make a commit on my behalf.

cbush: I put the reminder to commit in the contrib/README.md. Thoughts?

contrib/update_browser_readme_version.sh
```diff
@@ -0,0 +1,34 @@
...
+  ../packages/browser/sdk/README.md
+
+STATUS="$(git status -s)"
+if [ ! -n "$STATUS" ]; then
+  echo "Nothing to commit."
+  exit 0
+fi
+
+echo "README updated."
+git add -u "../packages/browser/sdk/README.md"
```
adamchel: [q] why `git add -u`? we don't want to add all changes

cbush: -u is just update which means it won't add that file if it wasn't already tracked. Not really necessary I suppose but it doesn't hurt either.

PR #258, opened 2019-04-01
STITCH-2426 Add Draft endpoints and tests, drive-by for some secret test names

marcgurney:
quick one, will commit the dist files once it's ready to merge


test/admin/draft.test.js
```diff
@@ -0,0 +1,41 @@
...
+  beforeEach(async() => {
+    const { apiKey, groupId, serverUrl } = extractTestFixtureDataPoints(test);
+    th = await buildAdminTestHarness(true, apiKey, groupId, serverUrl);
+    appDraft = th.app().draft();
+  });
+
+  afterEach(async() => th.cleanup());
+
+  it('responds with a 404 when attempting to get a draft that does not exist', async() => {
+    try {
```
mpobrien: should use `expect.assertions` in code blocks like this one. otherwise if it *doesn't* 404, then the `catch` block never executes and the assertions never get hit, so the test will silently pass even though the code didn't do what it's supposed to do.

test/admin/draft.test.js
```diff
@@ -0,0 +1,41 @@
...
+
+  beforeEach(async() => {
+    const { apiKey, groupId, serverUrl } = extractTestFixtureDataPoints(test);
+    th = await buildAdminTestHarness(true, apiKey, groupId, serverUrl);
+    appDraft = th.app().draft();
+  });
+
+  afterEach(async() => th.cleanup());
+
+  it('responds with a 404 when attempting to get a draft that does not exist', async() => {
```
mpobrien: it doesn't look like we have a test for `.get()` in the non-404 case (when the draft does exist) - maybe worth adding that too?

marcgurney: oops, duh, totally missed that

mpobrien:
LGTM when it's green!

PR #268, opened 2019-04-15
STITCH-2781 - Expose credentials options to get

edaniels:


edaniels:
cc: @marcgurney this is how we get the credentials to be set via the SDK

src/common.js
```diff
@@ -32,12 +32,18 @@ export const checkStatus = (response) => {
...
 
-export const makeFetchArgs = (method, body) => {
+export const makeFetchArgs = (method, body, options) => {
   const init = {
     method: method,
     headers: { 'Accept': JSONTYPE, 'Content-Type': JSONTYPE }
   };
 
+  if (options) {
+    if (options.credentials) {
```
haleyowen: I think you can combine these two if statements `if (options && options.credentials)`

PR #274, opened 2019-04-25
STITCH-2434/5 Add deploy config GET and PATCH endpoint handlers

marcgurney:


test/admin/deploy.test.js
```diff
@@ -18,8 +18,31 @@ describe('Deploy', () => {
...
-    expect(deployHistory).toEqual([]);
+  describe('when requesting the deploy history', () => {
+    it('returns a list containing the initial deploy', async() => {
+      const deployHistory = await appDeploy.deployments().list();
+      expect(deployHistory[0].status).toEqual('successful');
+    });
+  });
+
+  describe('when requesting the deploy config', () => {
+    it('returns the current config with an empty list of authenticated_repositories', async() => {
```
mpobrien: nit: i think you should have a space between `async` and the parens. Maybe run prettier on this file?

PR #275, opened 2019-05-09
STITCH-2914 - Add new endpoint for getting deployment installation

joshmadewell:
Adds a new endpoint to retrieve the deployment installation (which is configured once a user has linked their GitHub account to Stitch). Fixes a test for fetching deploy config which no longer returns the list of authenticated repositories.


test/admin/deploy.test.js
```diff
@@ -28,7 +28,7 @@ describe('Deploy', () => {
   describe('when requesting the deploy config', () => {
     it('returns the current config with an empty list of authenticated_repositories', async() => {
```
chickenbug: this description might need to change with this update

PR #282, opened 2019-06-10
Trigger onUserLoggedIn() event on OAuth login

tkaye407:


packages/core/sdk/src/auth/internal/CoreStitchAuth.ts
```diff
@@ -875,7 +875,12 @@ export default abstract class CoreStitchAuth<TStitchUser extends CoreStitchUser>
         // Dispatch a UserAdded event if this is the first time this user is 
         // Being added to the list of users on the device.
         if (newUserAdded) {
-          this.onAuthEvent() // Tegacy event dispatch
+          this.onAuthEvent() // Legacy event dispatch
+          
+          this.dispatchAuthEvent({
```
adamchel: [minor] this is actually not the correct place to be dispatching this event. We're already dispatching a login event for most of the login code paths in "doLogin".

See https://github.com/mongodb/stitch-js-sdk/blob/master/packages/core/sdk/src/auth/internal/CoreStitchAuth.ts#L720

The result is that on most logins that are happening via `doLogin`, you'll get two UserLoggedIn events, which is not the correct behavior.

You instead want to modify the code at https://github.com/mongodb/stitch-js-sdk/blob/master/packages/core/sdk/src/auth/internal/CoreStitchAuth.ts#L386 to dispatch a `UserLoggedIn` in the case of a successful result of `processLogin`.

PR #292, opened 2019-07-10
STITCH-3217: add admin sdk functions for confirming and listing unconfirmed users

dorinaaaaalin:


src/admin.js
```diff
@@ -292,7 +292,9 @@ export class StitchAdminClient extends StitchClient {
           }),
 
           userRegistrations: () => ({
-            sendConfirmationEmail: (email) => api._post(`${appUrl}/user_registrations/by_email/${email}/send_confirm`)
+            sendConfirmationEmail: (email) => api._post(`${appUrl}/user_registrations/by_email/${email}/send_confirm`),
+            confirmEmail: (email) => api._post(`${appUrl}/user_registrations/by_email/${email}/confirm`),
```
mburdette: this should probably be named `confirmByEmail`

src/admin.js
```diff
@@ -292,7 +292,9 @@ export class StitchAdminClient extends StitchClient {
...
+            unconfirmedUsersQuery: () => api._post(`${appUrl}/user_registrations/unconfirmed_users`)
```
mburdette: in general these should be readable, what you have here would be read as `userRegistrations.unconfirmedUsersQuery` which makes less sense than something like `userRegisrations.listUnconfirmed`

I think this also needs to take parameters to account for limit and afterid right?

test/auth.test.js
```diff
@@ -126,6 +145,27 @@ describe('Auth linking', () => {
...
     }
   });
+
+  it('should list pending users', async() => {
+    const testEmail = 'test_user@10gen.com';
+    await client.register(testEmail, password);
+    const userId = await client.login();
+    await client.register(linkEmail, password);
+
+    let { token_id: tokenId, token } = await th.app().userRegistrations().sendConfirmationEmail(linkEmail);
```
mburdette: I think you should just use your new `confirmByEmail` endpoint instead of doing the whole sending an email thing. Also you should add a test for when the limit is actually used e.g. more than 3 records the result should only contain 3 records.

test/auth.test.js
```diff
@@ -115,6 +116,24 @@ describe('Auth linking', () => {
     expect(newIdentities[1].provider_type).toEqual('local-userpass');
   });
 
+  it('should link two accounts without sending a confirmation email', async() => {
```
mburdette: this test string seems very confusing, it seems like this test is testing that you can link a user that was confirmed via `confirmByEmail` the string should more accurately represent that

Issue #209, opened 2019-01-30
watch method on RemoteMongoCollection in ReactNative


adrian-gierakowski:
Creating this issue as advised by @adamchel in #23

adamchel:
Hi @adrian-gierakowski thanks for filing this issue.

As per https://github.com/react-native-community/discussions-and-proposals/issues/99, we are planning on contributing support for server-eent events in React Native, which will allow us to implement this without resorting to a hack around `XmlHttpRequest`.

We hope to start work on this in the coming weeks!

adrian-gierakowski:
Hi @adamchel. Has there been any progress on this?

adamchel:
@adrian-gierakowski 

This work got re-prioritized, so I wasn't able to work on it as soon as I would have liked, but I've finally opened a PR on the official React Native repo, so hopefully we will be able to support this soon!

https://github.com/facebook/react-native/pull/25718

adrian-gierakowski:
That's great news @adamchel. Can't wait to try it out.

BrianIto:
Hello Adam! I'm Brian and can't wait to see it working. You did it in a Video in [this link](https://www.youtube.com/watch?v=-OoIpkAxmcw&feature=youtu.be) right?

It was great to the see! Any news please keep us on touch! Thank y'all for your work!

adamchel:
Hey everyone, we've just released experimental `watch` support for React Native! See https://github.com/mongodb/stitch-js-sdk/releases/tag/v4.9.0 for details on how you can use it. We can release this non-experimentally once React Native 0.62 is officially released.

If you have any problems with it, feel free to open a new issue!

PR #294, opened 2019-07-18
STITCH-2499 Add support for 'watch' in React Native

adamchel:
This adds support for "watch" functionality in React Native, as requested by https://github.com/mongodb/stitch-js-sdk/issues/209.

However, this change depends on https://github.com/facebook/react-native/pull/25718 getting accepted and released in a future release of React Native:  

I will be keeping this PR open until that happens.


BrianIto:
Hello everybody! I'm just a Junior Developer and just want to understand what's going on right now. React Native added support for the proper networking components to work?

adamchel:
This won't be supported until React Native 0.62 is released.

packages/react-native/core/src/core/internal/net/RNFetchStreamTransport.ts
```diff
@@ -51,6 +53,37 @@ import {
...
+        headers: reqHeaders,
+        method: request.method,
+        mode: 'cors'
+      }).then(response => {
+        const respHeaders: { [key: string]: string } = {};
+        response.headers.forEach((value, key) => {
+          respHeaders[key] = value;
+        });
+        if (response.status < 200 || response.status >= 300) {
+          return response.text()
```
tkaye407: Does response.text() return a promise?

packages/react-native/services/mongodb-remote/__tests__/RemoteMongoClientIntTests.ts
```diff
@@ -749,4 +753,262 @@ describe("RemoteMongoClient", () => {
 
     expect(coll.findOne()).rejects.toBeTruthy();
   });
+
+   /* tslint:disable:no-string-literal */
+   it("should properly support watch streams on set of ids", async () => {
```
tkaye407: I think you should split this up into smaller tests personally. This is essentially just 4 tests put into one test

adamchel: Discussed offline, these tests are basically just copy-pasted from the `watch` tests in the other two SDKs. We're going to leave as is.

packages/react-native/core/src/core/internal/net/RNFetchStreamTransport.ts
adamchel: yup

adamchel: from `lib.dom.d.ts`

```ts
interface Body {
    readonly body: ReadableStream<Uint8Array> | null;
    readonly bodyUsed: boolean;
    arrayBuffer(): Promise<ArrayBuffer>;
    blob(): Promise<Blob>;
    formData(): Promise<FormData>;
    json(): Promise<any>;
    text(): Promise<string>;
}
```

PR #297, opened 2019-07-30
STITCH-3299 - Add admin sdk method for deleting pending users

cktang88:


test/userRegistrations.test.js
```diff
@@ -0,0 +1,44 @@
...
+
+  afterEach(async() => th.cleanup());
+
+  it('removePendingUserByEmail should remove existing user', async() => {
+    await client.register(testEmail, password);
+    let users = await th.app().userRegistrations().listPending();
+    expect(users).toHaveLength(2);
+    await th.app().userRegistrations().removePendingUserByEmail(testEmail);
+    users = await th.app().userRegistrations().listPending();
+    expect(users).toHaveLength(1);
```
mburdette: you should actually check the users here to make sure the correct one was removed, can probably find some other assertion to do it or make a new list or something and do an equality check. 

https://jestjs.io/docs/en/expect

test/userRegistrations.test.js
```diff
@@ -0,0 +1,44 @@
...
+  });
+  it('removePendingUserByID should remove existing user', async() => {
+    await client.register(testEmail, password);
+    let users = await th.app().userRegistrations().listPending();
+    expect(users).toHaveLength(2);
+    await th.app().userRegistrations().removePendingUserByID(users[0]._id);
+    users = await th.app().userRegistrations().listPending();
+    expect(users).toHaveLength(1);
```
mburdette: you should actually check the users here to make sure the correct one was removed

test/userRegistrations.test.js
```diff
@@ -0,0 +1,44 @@
...
+
+  const testEmail = 'link_user@10gen.com';
+  const password = '123456';
+
+  beforeEach(async() => {
+    const { apiKey, groupId, serverUrl } = extractTestFixtureDataPoints(test);
+    th = await buildClientTestHarness(apiKey, groupId, serverUrl);
+
+    client = th.stitchClient;
+    // console.log(client);
```
mburdette: errant comment

test/userRegistrations.test.js
```diff
@@ -0,0 +1,44 @@
...
+    await client.logout();
+  });
+
+  afterEach(async() => th.cleanup());
+
+  it('removePendingUserByEmail should remove existing user', async() => {
```
mburdette: `should remove an existing pending user`

test/userRegistrations.test.js
```diff
@@ -0,0 +1,44 @@
...
+    await client.register(testEmail, password);
+    let users = await th.app().userRegistrations().listPending();
+    expect(users).toHaveLength(2);
+    await th.app().userRegistrations().removePendingUserByEmail(testEmail);
+    users = await th.app().userRegistrations().listPending();
+    expect(users).toHaveLength(1);
+  });
+  it('removePendingUserByID should remove existing user', async() => {
```
mburdette: `should remove an existing pending user`

dist/node/admin.js
```diff
@@ -501,8 +501,14 @@ var StitchAdminClient = exports.StitchAdminClient = function (_StitchClient) {
                 confirmByEmail: function confirmByEmail(email) {
                   return api._post(appUrl + '/user_registrations/by_email/' + email + '/confirm');
                 },
-                listUnconfirmed: function listUnconfirmed(filter) {
-                  return api._get(appUrl + '/user_registrations/unconfirmed_users', filter);
+                listPending: function listPending(filter) {
```
mburdette: i think you need to rebuild and then push, this still says `listUnconfirmed` from before, which is incorrect

cktang88: the diff seems to say listPending replaces listUnconfirmed in this file now?

mburdette: i guess someone forgot to rebuild

PR #299, opened 2019-08-06
get support branch closer to green

mpobrien:


tkaye407:
Is this what martin was asking me about?

.evg.yml
```diff
@@ -124,13 +124,21 @@ tasks:
...
+
             chown -R $USER stitch-js-sdk
             cd stitch-js-sdk
             curl -OL https://yarnpkg.com/latest.tar.gz
             tar zvxf latest.tar.gz
             mv yarn-* yarn
             export PATH="`pwd`/yarn/bin:$PATH"
-            yarn install
+
+            npm install
```
mburdette: I think you could leave this as `yarn install` and just remove the `package-lock.json` and replace it with the resulting `yarn-lock` from a `yarn install`

mpobrien:
@tkaye407 yes (presumably). This fix doesn't make the tests green but it does at least get it past the `npm install` phase, so I'll probably fix this up and tag someone to review soon, then we can move on to fixing the actual tests.

Issue #186, opened 2018-12-23
 window is not defined 


xxyyzz2050:
I'm using babel transpiler

```
unction StitchAuthImpl(requestClient, browserAuthRoutes, authStorage, appInfo, jsdomWindow) {
        if (jsdomWindow === void 0) { jsdomWindow = window; }
}
```

//error:  window is not defined

jsflax:
Hi @xxyyzz2050 ,

Could you give us a bit more information? What step in the transpiration process are you at? What is the full stack trace?

Thanks.

maredudd:
@jsflax I imagine that this might be due to using the SDK for SSR. I'm having the same issue with Next.js. 

If fails on this line: `client.auth.loginWithCredential(new AnonymousCredential())`:

```
ReferenceError: window is not defined
    at new StitchAuthImpl (/xxx/node_modules/mongodb-stitch-browser-core/dist/cjs/core/auth/internal/StitchAuthImpl.js:28:39)
    at new StitchAppClientImpl (/xxx/node_modules/mongodb-stitch-browser-core/dist/cjs/core/internal/StitchAppClientImpl.js:15:21)
    at Function.Stitch.initializeAppClient (/xxx/node_modules/mongodb-stitch-browser-core/dist/cjs/core/Stitch.js:71:22)
```

This would be expected since `window` is undefined on the server side. I tried using the server SDK to retrieve the data on the server side but that fails since `fs` doesn't exist  This results in the following code:

```import {
  Stitch,
  RemoteMongoClient,
  AnonymousCredential
} from 'mongodb-stitch-server-sdk'```

showing an error due to the missing dependency:

```This dependency was not found:

* fs in ./node_modules/fs-extra/lib/index.js, ./node_modules/fs-extra/lib/empty/index.js and 2 others```

Are the SDKs expected to support SSR? Is there a workaround for this or is Stitch not an option with SSR?

Thanks.

agstover:
I've been trying to figure out how Stitch can be used with SSR as well, particularly NextJS. Would one have to instantiate both a Stitch-Server SDK instance and a Stitch-Browser SDK instance, use the server instance if !window and use the browser instance if there is?

[...]

PR #302, opened 2019-08-15
use older 2.0.4 version of whatwg-fetch in browser core package

neonguru:
This resolves issue #186 
The whatwg-fetch module is responsible for the window is not defined error.
By using the previous version of that module, the error does not occur and everything works as required.
Tested in my SSR application and it is working as intended.

### DIFF
--- a/packages/browser/core/package-lock.json
+++ b/packages/browser/core/package-lock.json
@@ -1,11 +1,19 @@
 {
-  "requires": true,
+  "name": "mongodb-stitch-browser-core",
+  "version": "4.5.0",
   "lockfileVersion": 1,
+  "requires": true,
   "dependencies": {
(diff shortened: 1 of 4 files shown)


adamchel:
Haven't had the time to look at this too closely yet, but @neonguru do you have any idea what changed in `whatwg-fetch` between 2.0.4 and 3.0.0 that caused the `window is not defined` error? Would it perhaps be better to open an issue/PR with `whatwg-fetch` itself?

adamchel:
I'm hesitant to downgrade a dependency without understanding what changed.

neonguru:
There is an issue in their repo already regarding this:
https://github.com/github/fetch/issues/657

neonguru:
But it's going on a year that they've known about it...

PR #319, opened 2019-10-22
STITCH-3376 add client sdk method for calling userpass reset function

jsflax:


packages/browser/coretest/__tests__/StitchAppClientIntTests.ts
```diff
@@ -519,4 +519,34 @@ describe("StitchAppClient", () => {
...
+        "http://emailConfirmUrl.com",
+        "http://resetPasswordUrl.com",
+        "email subject",
+        "password subject",
+        true,
+        stitchFunction.id,
+        stitchFunction.name
+      ))
+    );
+  });
```
mburdette: shouldn't this test also actually reset the password? and then try to login with the new password?

packages/browser/coretest/__tests__/StitchAppClientIntTests.ts
```diff
@@ -519,4 +519,34 @@ describe("StitchAppClient", () => {
...
+      undefined,
+    ));
+
+    await harness.addProvider(
+      app,
+      new UserpassProvider(new UserpassProviderConfig(
+        "http://emailConfirmUrl.com",
+        "http://resetPasswordUrl.com",
+        "email subject",
+        "password subject",
```
mburdette: if you are resetting the password with a function you don't need a url or subject for the password stuff

packages/core/admin-client/src/authProviders/ProviderConfigs.ts
```diff
@@ -49,7 +49,13 @@ class UserpassProviderConfig {
...
     @jsonProperty("confirmEmailSubject") public confirmEmailSubject: string,
     
-    @jsonProperty("resetPasswordSubject") public resetPasswordSubject: string
+    @jsonProperty("resetPasswordSubject") public resetPasswordSubject: string,
+
+    @jsonProperty("runResetFunction") public runResetFunction = false,
+
+    @jsonProperty("resetFunctionId") public resetFunctionId = "",
+
+    @jsonProperty("resetFunctionName") public resetFunctionName = ""
```
mburdette: while you do this, you should probably add the fields for confirmation function as well 
```
"runConfirmationFunction"
"confirmationFunctionId"
"confirmationFunctionName"
```


https://github.com/10gen/stitch/blob/a0a456ed1189ea3a6fecdf6e40f222a1f43618df/authprovider/providers/local/userpass.go#L76

PR #349, opened 2020-01-02
STITCH-4077 Javascript SDK: Decode and expose `customData` on UserProfile

jsflax:


packages/browser/coretest/__tests__/StitchCustomUserDataIntTests.ts
```diff
@@ -0,0 +1,81 @@
...
+describe("StitchAppClient", () => {
+	const db = "db1"
+	const coll = "coll1"
+	const functionName = "addUserProfile";
+	const user = "stitch@10gen.com";
+	const password = "stitchuser";
+	const colorField = "favoriteColor";
+	const favoriteColor = "blue";
+
+	it("should define custom data", async () => {
```
mburdette: I feel like this test structure and string is a bit off. should there be a nested `describe` like `with custom user data` then you have one `it` block `user should have custom data` then another for like `user should have updated custom data after running refresh`? Instead of having everything in one big block. If you don't want to split this up better you could at least make the `it` string a little more descriptive of what is going on.

packages/core/admin-client/package.json
```diff
@@ -63,7 +63,7 @@
     "build-main": "tsc -p tsconfig.cjs.json",
     "build-module": "tsc -p tsconfig.esm.json",
     "prepare": "npm run build",
-    "test": "jest",
+
```
mburdette: [nit] remove this gap?

PR #354, opened 2020-01-13
STITCH-4374 - add Realm Config support

chickenbug:
Added the get and put actions for realm config


chickenbug:
Note: tests should fail until the server PR goes in

src/admin.js
```diff
@@ -447,6 +447,16 @@ export class StitchAdminClient extends StitchClient {
...
               validate: () => api._get(`${graphqlUrl}/validate`)
             };
+          },
+
+          realm: () => {
+            const realmUrl = `${appUrl}/realm`;
+            return {
+              config: () => ({
+                get: () => api._get(`${realmUrl}/config`),
+                update: (data) => api._put(realmUrl, { body: JSON.stringify(data) })
```
malexandert: Oh wait actually, both of these hit the `/config` endpoint, right? So it should also be `${realmUrl}/config`. Or, even better, we could just revise the `const` to be
```
realmConfigUrl = `${appUrl}/realm/config`
```
and then reference that in `get` and `update`

PR #356, opened 2020-01-16
STITCH-4381: Add schema generation support to admin sdk

malexandert:
So a notable omission here is the success case test. Even with the backend code in, that will still fail until we implement generation for at least one language


src/admin.js
```diff
@@ -458,6 +458,12 @@ export class StitchAdminClient extends StitchClient {
                   get: () => api._get(realmConfigUrl),
                   update: (data) => api._put(realmConfigUrl, { body: JSON.stringify(data) })
                 };
+              },
+              clientSchemas: () => {
+                const realmClientSchemasUrl = `${realmUrl}/client_schemas`;
+                return {
+                  get: language => api._get(`${realmClientSchemasUrl}/${language}`)
```
chickenbug: This needs to set the body as well using `data` to handle namespaces. You'll need to figure out if we should do it like `_put` or `_post` though.

malexandert: Discussed offline, but we're gonna update the route to support namespaces in the query params as opposed to the body, since `node-fetch` doesn't allow bodies with GET requests

chickenbug: We should update this to have (language, filter) =>

PR #367, opened 2020-02-24
STITCH-4698 - change realm sdk methods to sync and add schema gen success test

chickenbug:
Since we now have sync config/methods under sync instead of realm we need to change the sdk to support it

Also now that we have language support, we can add a sync schema gen success test for the sdk

### DIFF
--- a/test/admin/sync.test.js
+++ b/test/admin/sync.test.js
@@ -1,12 +1,12 @@
 const StitchMongoFixture = require('../fixtures/stitch_mongo_fixture');
 
-import { buildAdminTestHarness, extractTestFixtureDataPoints, createSampleMongodbService } from '../testutil';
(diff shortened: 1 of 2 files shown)


test/admin/sync.test.js
```diff
@@ -66,12 +66,21 @@ describe('Realm', () => {
...
         }
       });
+
+      it('should succed with a supported language and and rules', async() => {
+        // Setup MongoDB service
+        const mongoDBService = await createSampleMongodbService(services);
+        await addRuleToMongodbService(services, mongoDBService, {database: 'foo', collection: 'bar'});
+        const schemas =  await syncClientSchemas.get('KOTLIN');
+        expect(schemas.length).toEqual(1);
+        console.log(schemas);
```
malexandert: [nit] leftovers?

PR #384, opened 2020-04-10
REALMC-5151: Replace mongodb-extjson with bson package 

malexandert:
The [mongodb-extjson](https://github.com/mongodb-js/mongodb-extjson) package is causing problems with mms' move to Node 12, so we're replacing it the [bson](https://github.com/mongodb/js-bson) package that supports all the same stuff


dist/node/admin.js
```diff
@@ -793,11 +796,11 @@ var StitchAdminClient = exports.StitchAdminClient = function (_StitchClient) {
               };
             },
 
-            realm: function realm() {
-              var realmUrl = appUrl + '/realm';
+            sync: function sync() {
+              var syncUrl = appUrl + '/sync';
```
mburdette, line 800: is this intentional?

mburdette:
@malexandert did you install this with your local ui and run all the tests to verify that it doesn't break anything

dist/node/admin.js
malexandert: Yeah this got changed [a little while ago](https://github.com/mongodb/stitch-js-sdk/commit/acde8cdae6d6dcde8a8c76cbef142580a20827de#diff-75d19e3da4632ee1f6cc2cfd233abef0)

PR #408, opened 2020-05-13
REALMC-5421: Add get branches route

haleyowen:


src/admin.js
```diff
@@ -256,6 +256,13 @@ export class StitchAdminClient extends StitchClient {
               get: commit => api._get(`${appUrl}/deployments/${commit}`)
             }),
             installation: () => api._get(`${appUrl}/deploy/installation`),
+            repositories: repoId => ({
```
joshmadewell, line 259: this should probably be:

`repositories.repository(id).github().branches().list()`

```
repositories: () => ({
  repository: (repoId) => ({
    github: () => ({
      branches: () => ({
        list: () => ({
          return api._get...    
        }),
      }),
    }),
  });
}),
```

haleyowen: I thought we did the other one this way so that if we added another provider like bitbucket we could just add it to the end instead of copying all of it again? I was trying to keep this consistent with the auth route

PR #412, opened 2020-05-19
REALMC-4874 - add sync progress to admin client

chickenbug:


test/admin/sync.test.js
```diff
@@ -225,40 +225,6 @@ describe('Sync', () => {
...
-      try {
-        await addRuleToMongodbService(services, syncService, schemaInvalidatingRule );
-      } catch (e) {
-        stitchError = e;
-      }
-      expect(stitchError.code).toBe('InvalidSyncSchema');
-
-      const createResult = await addRuleToMongodbService(services, syncService, schemaInvalidatingRule, { allow_destructive_changes: true });
-      expect(createResult).toBeTruthy();
-    });
```
chickenbug, line 260: Things have changed in the server so you can no longer have a destructive change on create. Newly created invalid schemas are not destructive to sync as they are now ignored by sync entirely.

adamchel: Perhaps you should keep the test in and just remove the piece that expects an error and tries to bypass it. That way this confirms that you can add an sync-incompatible schema without experiencing an error.

test/admin/sync.test.js
```diff
@@ -225,38 +225,25 @@ describe('Sync', () => {
       expect(deployResult.status).toBe('created');
     });
 
-
-    it('can use the allow_destructive_changes param to do a destructive rule create', async() => {
+    it('can use do a create a rule with invalid sync schema', async() => {
```
andrewzlchen, line 228: Not sure if this was intentional, but the phrasing of the test name is kinda awkward.

Maybe we can change it to "can create a rule with invalid sync schema", or if it was intentional, add quotes to change it to "can do a 'Create a rule' with invalid sync schema"

chickenbug: Ah man, yeah this was a typo. Thanks!

PR #414, opened 2020-05-21
REALMC-5717: Allow creating temp api keys

Gabri3l:
This PR comes with some docs to help local development. I'm not sure if that was the right place to add some docs about `yalc`, let me know if I should move them somewhere else.


src/client.js
```diff
@@ -1,13 +1,13 @@
 /* global window, fetch */
```
Gabri3l, line 1: No relevant changes here, just my linter on save.

PR #417, opened 2020-05-27
REALMC-5801: Fix failing test on client sdk for mongodb service

Gabri3l:
This fixes the test but I want to make sure it makes sense since it only broke recently. I'm not sure if the order has importance here.

### DIFF
--- a/test/client/mongodb_service.test.js
+++ b/test/client/mongodb_service.test.js
@@ -297,8 +297,8 @@ describe('Client API executing mongodb service functions', () => {
     ]);
 
     expect(results).toMatchObject([
-      { _id: 'bar', total: 2 },
-      { _id: 'foo', total: 4 }
+      { _id: 'foo', total: 4 },
+      { _id: 'bar', total: 2 }
     ]);
   });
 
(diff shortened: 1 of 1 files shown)


test/client/mongodb_service.test.js
```diff
@@ -297,8 +297,8 @@ describe('Client API executing mongodb service functions', () => {
     ]);
 
     expect(results).toMatchObject([
-      { _id: 'bar', total: 2 },
```
tsedgwick, line 300: interesting - do we know why this order changed?

Gabri3l: Not sure to be honest, I haven't touched the client much to be honest. Maybe @makesitgo might have an idea ?

makesitgo: according to docs, [`$group` does not do any ordering](https://docs.mongodb.com/manual/reference/operator/aggregation/group/#pipe._S_group) ... seems like this was always fluky slash we were just getting lucky.

i'd recommend maybe just adding `{ $sort: { _id: 1 } }` to the pipeline, instead of reversing the order of the expected match object's fields

Gabri3l: Thank you, will do!

PR #423, opened 2020-06-04
REALMC-4900 - use new baas format

edaniels:


src/auth/common.js
```diff
@@ -1,12 +1,12 @@
-export const USER_AUTH_KEY = '_stitch_ua';
-export const REFRESH_TOKEN_KEY = '_stitch_rt';
-export const DEVICE_ID_KEY = '_stitch_did';
-export const STATE_KEY = '_stitch_state';
-export const USER_AUTH_COOKIE_NAME = 'stitch_ua';
-export const STITCH_ERROR_KEY = '_stitch_error';
-export const STITCH_LINK_KEY = '_stitch_link';
-export const USER_LOGGED_IN_PT_KEY = '_stitch_pt';
-export const STITCH_REDIRECT_PROVIDER = '_stitch_rp';
+export const USER_AUTH_KEY = '_baas_ua';
```
makesitgo, line 1: can't comment on the file, but I think you also want an update here: https://github.com/edaniels/stitch-js-sdk/blob/REALMC-4900/src/auth/storage.js#L84 (update L#34 in that file as well).

---

[ocd] the way some of these things get stored, we end up with double stamping of `stitch`/`baas`...

image

I'm having a hard time deciding if this would be a breaking change or not, but maybe we could just remove the `_baas_` part of these key strings all together?  that or re-work the way we "namespace" these keys from the Storage implementation

edaniels: I'd like to keep these namespaced but will address

PR #432, opened 2020-08-17
REALMC-6613 revert the libmongo url

arahmanan:


arahmanan:
The evergreen failures are not related to this change. Going ahead and merging.

PR #442, opened 2021-07-08
REALMC-9230: fix failing test in 'stitch-js-sdk-3.0'

nathantfrank:
The app cache was not being invalidated because the test setup requests were coming in too quickly.


kpatel71716 (approved):
LGTM! To add to the above comment, this should stop the tests from being flakey since the deployment from adding the service is now guaranteed to not be during the same timestamp that the app is created

tsedgwick (approved):
🙌

